PR #668, opened 2019-07-15
IBP-2872 Remove Suffix Column from Key_Sequence_register table

mderamos:
Kindly review. Thanks!


src/main/resources/liquibase/crop_changelog/12_5_0.xml
```diff
@@ -0,0 +1,80 @@
...
+                AND ksr.last_used_sequence > duplicate.last_used_sequence;
+		</sql>
+	</changeSet>
+
+	<changeSet author="mderamos" id="v12.5.0-2">
+		<preConditions onFail="MARK_RAN">
+			<columnExists tableName="key_sequence_register" columnName="suffix"/>
+		</preConditions>
+		<comment>
+			Remove suffix column from key_sequence_register table
```
darla-leafnode: As discussed, please perform dropping of unique index first before dropping the suffix column

src/main/resources/liquibase/crop_changelog/12_5_0.xml
```diff
@@ -0,0 +1,80 @@
...
+							AND ksr.last_used_sequence > duplicate.last_used_sequence;
+					</sqlCheck>
+				</not>
+			</and>
+		</preConditions>
+		<comment>
+			Remove duplicate entries from key_sequence_register
+		</comment>
+		<sql dbms="mysql" splitStatements="true">
+			DELETE duplicate FROM key_sequence_register duplicate, key_sequence_register ksr
```
darla-leafnode: We need to handle cases of duplicate prefixes with the same `last_used_sequence `value

PR #672, opened 2019-07-30
IBP-2784 Deprecate crop users and persons tables

abatac:


src/main/resources/liquibase/workbench_changelog/12_4_0.xml
```diff
@@ -0,0 +1,65 @@
...
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<tableExists tableName="crop_persons" />
+			</not>
+		</preConditions>
+		<comment>
+			Create CROP_PERSONS table for storing association between crop and persons
+		</comment>
+		<sql dbms="mysql" splitStatements="true">
+			CREATE TABLE crop_persons (
```
nahuel-soldevilla: We'll need to remove `users_crop` in the roles branch and change the code to use this table instead if I'm not mistaken. At least with this approach we'll have less conflicts while merging the branches.

cc @darla-leafnode @clarysabel

src/main/java/org/generationcp/middleware/manager/WorkbenchDaoFactory.java
```diff
@@ -0,0 +1,108 @@
...
+import org.generationcp.middleware.dao.RoleDAO;
+import org.generationcp.middleware.dao.StandardPresetDAO;
+import org.generationcp.middleware.dao.ToolDAO;
+import org.generationcp.middleware.dao.UserInfoDAO;
+import org.generationcp.middleware.dao.WorkbenchSidebarCategoryDAO;
+import org.generationcp.middleware.dao.WorkbenchSidebarCategoryLinkDAO;
+import org.generationcp.middleware.dao.WorkbenchUserDAO;
+import org.generationcp.middleware.hibernate.HibernateSessionProvider;
+
+public class WorkbenchDaoFactory {
```
nahuel-soldevilla: We'll need to migrate some methods from the DaoFactory here. I guess it's less confusing to have them here, you can identify quickly on which db you are working on

src/main/java/org/generationcp/middleware/pojos/workbench/CropPersonId.java
```diff
@@ -0,0 +1,53 @@
+package org.generationcp.middleware.pojos.workbench;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.generationcp.middleware.pojos.Person;
+
+import java.io.Serializable;
+
+public class CropPersonId implements Serializable {
```
nahuel-soldevilla: Why not CropPersonDTO?

src/main/java/org/generationcp/middleware/service/api/user/UserService.java
```diff
@@ -0,0 +1,365 @@
...
+import org.generationcp.middleware.pojos.workbench.ProjectUserInfo;
+import org.generationcp.middleware.pojos.workbench.Role;
+import org.generationcp.middleware.pojos.workbench.UserInfo;
+import org.generationcp.middleware.pojos.workbench.WorkbenchUser;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public interface UserService {
```
nahuel-soldevilla: Note to self: manually merge with roles branch service of the same name

src/main/java/org/generationcp/middleware/pojos/workbench/CropPersonId.java
abatac: This is actually used as a Hibernate composite primary key class for CropPerson.

PR #676, opened 2019-08-14
IBP-2635 export csv improvements

mderamos:
Kindly review. Thanks!

### DIFF
--- a/src/main/java/org/generationcp/middleware/dao/dms/GeolocationPropertyDao.java
+++ b/src/main/java/org/generationcp/middleware/dao/dms/GeolocationPropertyDao.java
@@ -13,6 +13,7 @@ package org.generationcp.middleware.dao.dms;
 
 import com.google.common.base.Preconditions;
 import org.generationcp.middleware.dao.GenericDAO;
+import org.generationcp.middleware.domain.oms.TermId;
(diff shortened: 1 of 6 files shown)


src/main/java/org/generationcp/middleware/dao/dms/GeolocationPropertyDao.java
```diff
@@ -139,6 +139,29 @@ private void deleteValues(final int studyId,final int termId) {
 		}
 	}
 
+	public Map<Integer, String> getInstanceIdLocationIdMap(final List<Integer> instanceIds) {
+		Map<Integer, String> instanceIdLocationIdMap = new HashMap<>();
+		final StringBuilder sql =
+			new StringBuilder().append("SELECT  ").append("    geo.nd_geolocation_id as instanceId, geo.value as value ").append("FROM ")
+				.append(" nd_geolocationprop geo ").append("WHERE ").append("    geo.nd_geolocation_id in (:geolocationIds) ")
+				.append("        AND geo.type_id = 8190");
```
darla-leafnode: Use `query.setParameter` and use `TermId.LOCATION_ID.getId() ` instead of appending to query string

PR #686, opened 2019-09-06
IBP-2946 Fix Import Dataset

abatac:


src/main/java/org/generationcp/middleware/operation/parser/WorkbookParser.java
```diff
@@ -443,105 +421,88 @@ protected Date validateDate(final String dateString, final boolean isStartDate,
...
+			this.assignVariableTypeAndRoleBasedOnSectionName(sectionName, measurementVariable);
+
+			measurementVariables.add(measurementVariable);
+
+			this.rowIndex++;
+		}
+	}
+
+	protected void assignVariableTypeAndRoleBasedOnSectionName(final String sectionName, final MeasurementVariable measurementVariable) {
+		if (Section.STUDY_DETAILS.getName().equals(sectionName)) {
```
darla-leafnode: I would role, label and VariableType as properties of Section enum to avoid this series of if-else statements

PR #691, opened 2019-09-30
Change if trndate from int to Date

vmaletta:
Please merge

### DIFF
--- a/src/test/java/org/generationcp/middleware/data/initializer/InventoryDetailsTestDataInitializer.java
+++ b/src/test/java/org/generationcp/middleware/data/initializer/InventoryDetailsTestDataInitializer.java
@@ -1,11 +1,6 @@
 
 package org.generationcp.middleware.data.initializer;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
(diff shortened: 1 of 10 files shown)


src/test/java/org/generationcp/middleware/data/initializer/InventoryDetailsTestDataInitializer.java
```diff
@@ -18,22 +13,36 @@
...
 	private static final String GERMPLASM_ENTITY_TYPE = "GERMPLSM";
 	private static final int USER_ID = 1;
 	private static final int NO_OF_STOCK_LIST_ENTRIES = 20;
-
-	public static final Integer DATE = 19122016;
 	public static final String LIST_NAME = "List1";
 	public static final String USER = "User";
 	public static final String STATUS = "Active";
 	public static final Double AMOUNT = -50.0;
+	public Date DATE;
```
clarysabel: @vmaletta if it is not a constant, it should not be uppercase

vmaletta: Got it!

PR #693, opened 2019-10-08
Ibp 2976 move stock

vmaletta:
Please merge

### DIFF
--- a/src/main/java/org/generationcp/middleware/pojos/ims/Transaction.java
+++ b/src/main/java/org/generationcp/middleware/pojos/ims/Transaction.java
@@ -125,7 +125,10 @@ public class Transaction implements Serializable {
 		this.sourceRecordId = sourceRecordId;
 		this.previousAmount = previousAmount;
 		this.personId = personId;
-		this.lot.setStockId(inventoryID);
+
+		if (this.lot != null) {
+			this.lot.setStockId(inventoryID);
+		}
 	}
 
 	public Integer getId() {
--- a/src/main/java/org/generationcp/middleware/pojos/ims/Transaction.java
+++ b/src/main/java/org/generationcp/middleware/pojos/ims/Transaction.java
@@ -44,8 +44,6 @@ public class Transaction implements Serializable {
 	private static final long serialVersionUID = 77866453513905521L;
 	public static final String GET_EMPTY_LOT = "getEmptyLot";
 	public static final String GET_LOT_WITH_MINIMUM_AMOUNT = "getLotWithMinimumAmount";
-	public static final String GET_INVENTORY_ID_WITH_IDENTIFIER_QUERY = "select inventory_id FROM ims_transaction WHERE inventory_id "
-			+ "RLIKE '^:identifier[0-9][0-9]*.*'";
 
 	@Id
 	@GeneratedValue(strategy = GenerationType.IDENTITY)
@@ -95,9 +93,6 @@ public class Transaction implements Serializable {
 	@Column(name = "personid")
 	private Integer personId;
 
-	@Column(name = "inventory_id")
-	private String inventoryID;
-
 	@Column(name = "bulk_with")
 	private String bulkWith;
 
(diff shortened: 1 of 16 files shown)


src/main/resources/liquibase/crop_changelog/13_2_0.xml
```diff
@@ -486,5 +486,74 @@
 		</sql>
 	</changeSet>
 
+	<changeSet author="vanina" id="v13.2.0-1">
+		<preConditions onFail="MARK_RAN">
+			<not>
```
clarysabel: @vmaletta we could add that also exists ims_transaction.inventory_id

src/main/resources/liquibase/crop_changelog/13_2_0.xml
```diff
@@ -486,5 +486,74 @@
...
+			GROUP BY trx.lotid) AS t ON t.lotid = lot.lotid
+			SET lot.stock_id = t.inventory_id;
+		</sql>
+	</changeSet>
+
+	<changeSet author="vanina" id="v13.2.0-4">
+		<preConditions onFail="MARK_RAN">
+			<columnExists tableName="ims_transaction" columnName="inventory_id"/>
+		</preConditions>
+		<comment>Drop role column inventory_id from ims_transaction table</comment>
```
clarysabel: remove role column from comment

src/main/java/org/generationcp/middleware/pojos/ims/Transaction.java
```diff
@@ -130,7 +125,7 @@ public Transaction(final Integer id, final Integer userId, final Lot lot, final
 		this.sourceRecordId = sourceRecordId;
 		this.previousAmount = previousAmount;
 		this.personId = personId;
-		this.inventoryID = inventoryID;
+		this.lot.setStockId(inventoryID);
```
clarysabel: not sure if this is needed but we should check if lot is null in case that it is nullable in ims_transaction...

src/main/resources/liquibase/crop_changelog/13_2_0.xml
```diff
@@ -486,5 +486,74 @@
...
+			WHERE
+			trx.inventory_id IS NOT NULL
+			GROUP BY trx.lotid) AS t ON t.lotid = lot.lotid
+			SET lot.stock_id = t.inventory_id;
+		</sql>
+	</changeSet>
+
+	<changeSet author="vanina" id="v13.2.0-4">
+		<preConditions onFail="MARK_RAN">
+			<columnExists tableName="ims_transaction" columnName="inventory_id"/>
```
clarysabel: maybe we should also check that stock_id exists in ims_lot.

src/main/resources/liquibase/crop_changelog/13_2_0.xml
```diff
@@ -486,5 +486,74 @@
...
 	</changeSet>
 
+	<changeSet author="vanina" id="v13.2.0-1">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<tableExists tableName="stock_id_bkp"></tableExists>
+			</not>
+		</preConditions>
+		<comment>
+			Create stockId backup table and migrate inventory_id values for more than 1 stock_id
```
clarysabel: Comment is wrong because we are storing all inventory_ids except by the null ones

src/main/resources/liquibase/crop_changelog/13_2_0.xml
vmaletta: Ok

src/main/resources/liquibase/crop_changelog/13_2_0.xml
vmaletta: Ok

src/main/java/org/generationcp/middleware/pojos/ims/Transaction.java
vmaletta: Ok

src/main/resources/liquibase/crop_changelog/13_2_0.xml
vmaletta: Ok

src/main/resources/liquibase/crop_changelog/13_2_0.xml
vmaletta: Ok

PR #695, opened 2019-10-17
IBP-3107 Move Experimental Design to BMSAPI

abatac:


src/main/java/org/generationcp/middleware/dao/dms/ProjectPropertyDao.java
```diff
@@ -279,6 +273,15 @@ public void deleteProjectVariables(final Integer projectId, final List<Integer>
 		query.executeUpdate();
 	}
 
+	public void deleteProjectVariablesByVariableTypes(final Integer projectId, final List<Integer> variableTypeIds) {
```
clarysabel: I think we should start standardizing our concepts and start using dataset instead of project whenever is possible. It makes sense to rename this by deleteDatasetVariablesByVariableTypes.

src/main/java/org/generationcp/middleware/service/api/study/generation/ExperimentDesignService.java
```diff
@@ -0,0 +1,17 @@
...
+import com.google.common.base.Optional;
+import org.generationcp.middleware.domain.etl.MeasurementVariable;
+import org.generationcp.middleware.pojos.workbench.CropType;
+import org.generationcp.middleware.service.api.dataset.ObservationUnitRow;
+
+import java.util.List;
+
+public interface ExperimentDesignService {
+
+	void deleteExperimentDesign(int studyId);
```
clarysabel: I suggest to rename it by deleteStudyExperimentDesign or deleteExperimentDesignByStudyId. Not clear what the parameter refers to given that in most of the cases when we delete something the received parameter is the id of the object that will be deleted.

src/main/java/org/generationcp/middleware/service/api/study/generation/ExperimentDesignService.java
```diff
@@ -0,0 +1,17 @@
...
+
+	void saveExperimentDesign(CropType crop, int studyId, List<MeasurementVariable> variables, List<ObservationUnitRow> rows);
+
+	Optional<Integer> getExperimentDesignTypeTermId(int studyId);
```
clarysabel: same comment about the name of this function.

src/main/java/org/generationcp/middleware/service/impl/study/generation/ExperimentDesignServiceImpl.java
```diff
@@ -0,0 +1,240 @@
...
+				project.setProjectId(projectId);
+				final String value = variable.getValue();
+				final ProjectProperty property =
+					new ProjectProperty(project, variableTypeId, value, rank, variableId, variable.getAlias());
+				this.daoFactory.getProjectPropertyDAO().save(property);
+
+				// FIXME Undo this duplicate saving in nd_geolocationprop (logged as part of IBP-3150)
+				if (isEnvironmentVariable) {
+					for (final Geolocation geolocation : geolocations) {
+						final GeolocationProperty geolocationProperty = new GeolocationProperty();
```
clarysabel: does it helps to have a GeolocationProperty constructor passing these parameters? I think it will be more legible.

src/main/java/org/generationcp/middleware/service/impl/study/generation/ExperimentModelGenerator.java
```diff
@@ -0,0 +1,90 @@
...
+			if (measurementVariable != null && ExperimentModelGenerator.EXPT_DESIGN_TYPES.contains(measurementVariable.getVariableType())) {
+				experimentProperties.add(this.createTrialDesignProperty(experimentModel, measurementVariable, unitData.getValue(), rank));
+				rank++;
+			}
+		}
+
+		return experimentProperties;
+	}
+
+	private ExperimentProperty createTrialDesignProperty(final ExperimentModel experimentModel,
```
clarysabel: this function will not add so much value if we create an ExperimentModel constructor

src/main/java/org/generationcp/middleware/service/impl/study/generation/StockModelGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+	private void addProperty(final StockModel stockModel, final StockProperty property) {
+		if (stockModel.getProperties() == null) {
+			stockModel.setProperties(new HashSet<StockProperty>());
+		}
+		property.setStock(stockModel);
+		stockModel.getProperties().add(property);
+	}
+
+	private StockProperty createProperty(final Integer variableId, final String value, final Integer rank) {
+		final StockProperty property = new StockProperty();
```
clarysabel: same here, we are creating these functions that are basically constructors

PR #700, opened 2019-11-12
First commit

vmaletta:
Please merge


src/main/java/org/generationcp/middleware/domain/germplasm/GermplasmDTO.java
```diff
@@ -32,8 +32,6 @@
 	private String germplasmGenus;
 	private String germplasmSpecies;
 	private String seedSource;
-	private String studyDbId;
-	private String trialName;
```
nahuel-soldevilla: @vmaletta I think we should return the deprecated fields until v2, because some clients may expect fields from v1.2 when then call /brapi/v1/

cc @clarysabel

vmaletta: We were discussing with @clarysabel and we decided to delete this fields, because this call is only for 1.3

PR #710, opened 2019-11-22
Ibp 3174 create lot

vmaletta:
Please merge


src/main/java/org/generationcp/middleware/dao/LocationDAO.java
```diff
@@ -281,6 +282,46 @@ public long countByCountry(final Country country) {
...
+
+		} catch (final HibernateException e) {
+			throw new MiddlewareQueryException(
+				this.getLogExceptionMessage(LocationDAO.GET_BY_TYPE, "types", String.valueOf(types), e.getMessage(),
+					LocationDAO.CLASS_NAME_LOCATION), e);
+		}
+		return locations;
+	}
+
+	public List<Location> getFavoriteLocationsByTypes(final Set<Integer> types, final List<Integer> locIds) {
```
clarysabel: @cuenyad what locIds means?

src/main/java/org/generationcp/middleware/dao/ims/TransactionDAO.java
```diff
@@ -741,8 +743,24 @@ private String formatDate(final Date date) {
...
 			return TransactionDto.class.getConstructor(Integer.class, String.class, String.class,  Double.class,
 			 String.class, Date.class, Integer.class, Integer.class, String.class, String.class, Integer.class, String.class, String.class);
-		} catch (NoSuchMethodException ex) {
+		} catch (final NoSuchMethodException ex) {
 			throw new RuntimeException(ex);
 		}
 	}
+
+
+	public Transaction saveTransaction(final TransactionDto transactionDto, final Lot lot) {
```
clarysabel: @vmaletta why the entire lot is needed? Does the transaction contain the lotId? In that case the lot param is not needed. Why the status is not a param?

src/main/java/org/generationcp/middleware/dao/LocationDAO.java
```diff
@@ -281,6 +282,33 @@ public long countByCountry(final Country country) {
...
+				criteria.add(Restrictions.in(LocationDAO.LOCID, locationIds));
+			}
+
+			criteria.add(
+				Restrictions.or(Restrictions.eq(LocationDAO.UNIQUE_ID, programUUID), Restrictions.isNull(LocationDAO.UNIQUE_ID)));
+			criteria.addOrder(Order.asc(LocationDAO.LNAME));
+			locations = criteria.list();
+
+		} catch (final HibernateException e) {
+			throw new MiddlewareQueryException(
```
clarysabel: @cuenyad please log the error
`LocationDAO.LOG.error(e.getMessage(), e);`

src/main/java/org/generationcp/middleware/dao/LocationDAO.java
```diff
@@ -281,6 +282,33 @@ public long countByCountry(final Country country) {
...
+			if (types != null && !types.isEmpty()) {
+				criteria.add(Restrictions.in(LocationDAO.LTYPE, types));
+			}
+
+			if (locationIds != null && !locationIds.isEmpty()) {
+				criteria.add(Restrictions.in(LocationDAO.LOCID, locationIds));
+			}
+
+			criteria.add(
+				Restrictions.or(Restrictions.eq(LocationDAO.UNIQUE_ID, programUUID), Restrictions.isNull(LocationDAO.UNIQUE_ID)));
```
clarysabel: @cuenyad, does it work if programUUID is null?

src/main/java/org/generationcp/middleware/dao/LocationDAO.java
```diff
@@ -705,7 +733,7 @@ public Location saveOrUpdate(final Location location) {
 			locations = criteria.list();
 		} catch (final HibernateException e) {
 			throw new MiddlewareQueryException(
-					this.getLogExceptionMessage("getByIds", "", null, e.getMessage(), LocationDAO.CLASS_NAME_LOCATION), e);
+					this.getLogExceptionMessage("getByUniqueID", "programUUID", null, e.getMessage(), LocationDAO.CLASS_NAME_LOCATION), e);
```
clarysabel: @cuenyad I guess the function was not renamed, so it should still be `getByIds'

src/main/java/org/generationcp/middleware/dao/ims/TransactionDAO.java
```diff
@@ -741,8 +743,24 @@ private String formatDate(final Date date) {
...
 			 String.class, Date.class, Integer.class, Integer.class, String.class, String.class, Integer.class, String.class, String.class);
-		} catch (NoSuchMethodException ex) {
+		} catch (final NoSuchMethodException ex) {
 			throw new RuntimeException(ex);
 		}
 	}
+
+
+	public Transaction saveTransaction(final TransactionDto transactionDto, final Lot lot, final TransactionStatus transactionStatus) {
+		final Transaction transaction = new Transaction();
```
clarysabel: @vmaletta Lot should not be a parameter of this function if TransactionDto contains a Lot with the id. 
There is no need to query for the lot, just build a new Lot entity and set the id

clarysabel: @vmaletta if TransactionDto has a type, then the status can be calculated and not passed as a parameter to the function because it can cause some inconsistencies.

src/main/java/org/generationcp/middleware/dao/ims/TransactionDAO.java
```diff
@@ -741,8 +743,24 @@ private String formatDate(final Date date) {
...
+		transaction.setStatus(transactionStatus.getIntValue());
+		transaction.setLot(lot);
+		transaction.setPersonId(Integer.valueOf(transactionDto.getUser()));
+		transaction.setUserId(Integer.valueOf(transactionDto.getUser()));
+		transaction.setTransactionDate(new Date());
+		transaction.setQuantity(transactionDto.getAmount());
+		transaction.setPreviousAmount(0D);
+		transaction.setCommitmentDate(0);
+		transaction.setComments(transactionDto.getNotes());
```
clarysabel: @vmaletta not so sure about commitment date always being 0, I guess it depends on the TransactionType

src/main/java/org/generationcp/middleware/manager/LocationDataManagerImpl.java
```diff
@@ -513,4 +519,18 @@ public String retrieveLocIdOfUnspecifiedLocation() {
 		return unspecifiedLocationId;
 	}
 
+	@Override
+	public List<Integer> getFavoriteProjectLocationIds(final String programUUID) {
+		final List<ProgramFavorite> programFavorites =
+			this.getProgramFavoriteDao().getProgramFavorites(ProgramFavorite.FavoriteType.LOCATION, Integer.MAX_VALUE, programUUID);
```
clarysabel: @cuenyad please use DaoFactory to get programFavoriteDao

src/main/java/org/generationcp/middleware/manager/api/LocationDataManager.java
```diff
@@ -209,6 +210,18 @@
...
+	 * Returns the Location records with type matching the given parameter. The
+	 * data is retrieved from both local and central databases.
+	 *
+	 * @param types - search set for the types of locations
+	 * @param programUUID - unique id of the current program.
+	 * @param locationIds - List of favorite LocIds of the current program
+	 *
+  	 * @return List of Location POJOs
+	 */
+	List<Location> getFilteredLocations(final Set<Integer> types, final List<Integer> locationIds, final String programUUID);
```
clarysabel: @cuenyad remove final from interface signature

src/main/java/org/generationcp/middleware/manager/api/LocationDataManager.java
```diff
@@ -484,4 +497,14 @@
...
+
+
+	/**
+	 * Gets the favorite project location ids.
+	 *
+	 * @param programUUID
+	 *            - unique id of program
+	 * @return the favorite project location ids
+	 */
+	List<Integer> getFavoriteProjectLocationIds(final String programUUID);
```
clarysabel: @cuenyad remove final from interface signature

src/main/java/org/generationcp/middleware/service/impl/inventory/LotServiceImpl.java
```diff
@@ -25,12 +28,33 @@ public LotServiceImpl(final HibernateSessionProvider sessionProvider) {
...
 	}
 
 	@Override
 	public long countSearchLots(final LotsSearchDto lotsSearchDto) {
 		return this.daoFactory.getLotDao().countSearchLots(lotsSearchDto);
 	}
+
+	@Override
+	public Integer saveLot(final LotGeneratorInputDto lotDto) {
+		final String stockId = lotDto.getStockId();
```
clarysabel: @vmaletta having stockId variable is not needed anymore

src/main/java/org/generationcp/middleware/service/impl/inventory/LotServiceImpl.java
```diff
@@ -25,12 +28,33 @@ public LotServiceImpl(final HibernateSessionProvider sessionProvider) {
...
+		final Lot lot = new Lot();
+		lot.setUserId(lotDto.getUserId());
+		lot.setComments(lotDto.getComments());
+		lot.setCreatedDate(new Date());
+		lot.setEntityId(lotDto.getGid());
+		lot.setEntityType("GERMPLSM");
+		lot.setLocationId(lotDto.getLocationId());
+		lot.setStockId(stockId);
+		lot.setStatus(0);
+		lot.setSource(0);
```
clarysabel: @vmaletta not so sure about 0 being the source in all cases, we should be careful in the near future to set the right value here

src/main/java/org/generationcp/middleware/service/impl/inventory/LotServiceImpl.java
```diff
@@ -25,12 +28,33 @@ public LotServiceImpl(final HibernateSessionProvider sessionProvider) {
...
+		lot.setScaleId(lotDto.getScaleId());
+
+		this.daoFactory.getLotDao().save(lot);
```
clarysabel: @vmaletta are you creating the lot uuid? I think it is not.

src/main/java/org/generationcp/middleware/service/impl/inventory/TransactionServiceImpl.java
```diff
@@ -33,4 +34,11 @@ public TransactionServiceImpl(final HibernateSessionProvider sessionProvider) {
 	public long countSearchTransactions(final TransactionsSearchDto transactionsSearchDto) {
 		return this.daoFactory.getTransactionDAO().countSearchTransactions(transactionsSearchDto);
 	}
+
+	@Override
+	public Integer saveTransaction(final TransactionDto transactionDto, final TransactionStatus transactionStatus) {
```
clarysabel: TransactionStatus should be inferred based on the type

src/main/java/org/generationcp/middleware/dao/LocationDAO.java
cuenyad: @clarysabel the programUUID is mandatory for the service and the condition is to recover only the locations that have the same programUUID or the programUUID is null this is because you can create locations for a program.

src/main/java/org/generationcp/middleware/dao/LocationDAO.java
cuenyad: @clarysabel the name of the function is getByUniqueID not getByIds. The commit was to fixed that.

src/main/java/org/generationcp/middleware/dao/LocationDAO.java
cuenyad: @clarysabel I verified and if the programUUID is null the query continues working properly.

src/main/java/org/generationcp/middleware/dao/ims/TransactionDAO.java
```diff
@@ -741,8 +743,31 @@ private String formatDate(final Date date) {
...
+		} else {
+			transaction.setStatus(TransactionStatus.ANTICIPATED.getIntValue());
+		}
+		transaction.setLot(lot);
+		transaction.setPersonId(Integer.valueOf(transactionDto.getUser()));
+		transaction.setUserId(Integer.valueOf(transactionDto.getUser()));
+		transaction.setTransactionDate(new Date());
+		transaction.setQuantity(transactionDto.getAmount());
+		transaction.setPreviousAmount(0D);
+		//FIXME not so sure about commitment date always being 0, I guess it depends on the TransactionType
```
clarysabel: @vmaletta fix FIXME

src/main/java/org/generationcp/middleware/service/impl/inventory/LotServiceImpl.java
```diff
@@ -25,12 +35,33 @@ public LotServiceImpl(final HibernateSessionProvider sessionProvider) {
...
+		final Lot lot = new Lot();
+		lot.setUserId(lotDto.getUserId());
+		lot.setComments(lotDto.getComments());
+		lot.setCreatedDate(new Date());
+		lot.setEntityId(lotDto.getGid());
+		lot.setEntityType("GERMPLSM");
+		lot.setLocationId(lotDto.getLocationId());
+		lot.setStockId(lotDto.getStockId());
+		lot.setStatus(0);
+		//FIXME we should be careful in the near future to set the right value here
```
clarysabel: @vmaletta fix FIXME

PR #711, opened 2019-11-26
Ibp 3182 observation units brapi

vmaletta:
Please merge


src/main/java/org/generationcp/middleware/domain/search_request/ObservationUnitsSearchRequestDto.java
```diff
@@ -0,0 +1,101 @@
+package org.generationcp.middleware.domain.search_request;
```
clarysabel: we should have two packages? for v1 and v2?

src/main/java/org/generationcp/middleware/service/api/phenotype/PhenotypeSearchDTO.java
```diff
@@ -1,5 +1,9 @@
+
+
```
clarysabel: @vmaletta remove empty lines

src/main/java/org/generationcp/middleware/domain/search_request/ObservationUnitsSearchRequestDto.java
clarysabel: @vmaletta please reorganize the search requests in v1 and v2

vmaletta: ok

src/main/java/org/generationcp/middleware/service/api/phenotype/PhenotypeSearchDTO.java
vmaletta: ok

PR #713, opened 2019-12-02
IBP-3145 expt design mapping

mderamos:
Kindly review. Thanks

### DIFF
--- a/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
+++ b/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
@@ -325,6 +325,25 @@ public class DataImportServiceImpl extends Service implements DataImportService 
 	}
 
 	@Override
+	public void addExptDesignVariableIfNotExists(
+		final Workbook workbook, final List<MeasurementVariable> measurementVariables,
+		final String programUUID) {
+
+		final List<MeasurementVariable> combinedVariables = new ArrayList<>();
+		combinedVariables.addAll(workbook.getConditions());
+		combinedVariables.addAll(workbook.getFactors());
+
+		final Optional<MeasurementVariable> exptDesignExists = this.findMeasurementVariableByTermId(TermId.EXPERIMENT_DESIGN_FACTOR.getId(), combinedVariables);
+
+		if (!exptDesignExists.isPresent()) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -368,6 +392,57 @@ public void assignLocationIdVariableToEnvironmentDetailSection(final Workbook wo
...
+				}
+			} else {
+				exptDesignVar.setValue(this.getExperimentalDesignIdValue(exptDesignValueFromObsSheet));
+			}
+		}
+
+		final Optional<MeasurementVariable> exptDesignVarInFactors =
+			this.findMeasurementVariableByTermId(TermId.EXPERIMENT_DESIGN_FACTOR.getId(), workbook.getFactors());
+		if(exptDesignVarInFactors.isPresent()) {
+			final MeasurementVariable exptDesignVar = exptDesignVarInFactors.get();
```
abatac: Is there a case where exptDesignVarInFactors.get() is not present?

src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -368,6 +392,57 @@ public void assignLocationIdVariableToEnvironmentDetailSection(final Workbook wo
...
 
+	@Override
+	public void processExperimentalDesign(final Workbook workbook, final String programUUID, final String exptDesignValueFromObsSheet)
+		throws WorkbookParserException {
+		final Optional<MeasurementVariable> exptDesignVarInConditions =
+			this.findMeasurementVariableByTermId(TermId.EXPERIMENT_DESIGN_FACTOR.getId(), workbook.getConditions());
+		if(exptDesignVarInConditions.isPresent()) {
+			final MeasurementVariable exptDesignVar = exptDesignVarInConditions.get();
+			if (exptDesignVar.getValue() != null && !exptDesignVar.getValue().isEmpty()) {
+				Term exptDesignValue = this.termDataManager.getTermByName(exptDesignVar.getValue());
```
abatac: Is this the same code in getExperimentalDesignIdValue?

src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
mderamos: Yes. When the Expt_design is not present in the import file.

PR #715, opened 2019-12-05
IBP-3202 Add Visualization Tool

abatac:


src/main/java/org/generationcp/middleware/dao/dms/ObservationUnitsSearchDao.java
```diff
@@ -965,4 +1002,175 @@ private ObservationUnitRow getObservationUnitRow(final ObservationUnitsSearchDTO
...
+	/**
+	 * Stripped-down version of {@Link ObservationUnitsSearchDao#getObservationUnitTableQuery}. It will only return the values (no metadata -- PhenotypeId, experimentId, obs_unit_id, etc) from factors and variates at Observation level.
+	 * The query is also optimized to exclude columns that are not specified in {@Link ObservationUnitsSearchDTO#filter#filterColumns} parameter as to reduce the number of subqueries and ensure that the
+	 * query only returns columns that are needed per request.
+	 *
+	 * @param searchDto
+	 * @param observationUnitNoName
+	 * @return
+	 */
+	private String getObservationUnitTableAsListOfMapQuery(
```
darla-leafnode: As discussed: 
-it's essential for performance in the visualization option to optimize the query to only get information it needs (reduce subqueries etc)
-in the long run, it would be ideal to only maintain one query

darla-leafnode: We can maybe follow this logic wherever you are checking `filterColumns`:
`final boolean noFilterVariables = CollectionUtils.isEmpty(filterColumns);
		if (noFilterVariables || filterColumns.contains(TRIAL_INSTANCE)) {
			sqlColumns.add("    gl.description AS TRIAL_INSTANCE");
		}`

Just make sure to put a comment when filterColumns has a value and when it doesn't

src/main/java/org/generationcp/middleware/service/api/dataset/ObservationUnitsSearchDTO.java
```diff
@@ -41,16 +42,19 @@
...
 		public Filter() {
 			this.byMissing = false;
 			this.byOutOfBound = false;
-			this.byOutOfSync =  false;
+			this.byOutOfSync = false;
 			this.byOverwritten = false;
 			this.filteredValues = new HashMap<>();
 			this.filteredTextValues = new HashMap<>();
 			this.variableId = null;
+			/** This is used by Visualization tool, to specify the columns that will included in the data returned from the server. **/
```
darla-leafnode: It's a bit confusing to put `filterColumns` here because the others are filters for **rows** that will be included.
Maybe we should move it as variable of the main DTO class.

src/main/java/org/generationcp/middleware/dao/dms/ObservationUnitsSearchDao.java
```diff
@@ -28,9 +28,11 @@
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.StringJoiner;
```
darla-leafnode: StringJoiner is supported from Java 1.8 onwards. Middleware still builds on 1.7

src/main/java/org/generationcp/middleware/dao/dms/ObservationUnitsSearchDao.java
nahuel-soldevilla: @abatac 
I have just one question here: Did you experienced real performance degradation here with the all-columns query or was this decision only based on theoretical information? Does this new query improve the performance of the visualization significantly?

Thanks!

cc @darla-leafnode @clarysabel

abatac: Hi @nahuel-soldevilla, Yes, I did experience performance degradation using the all-columns query. Tested on a study with 30,000 records, when using the all-columns query, the generation of visualization graph takes around 2 min 30 seconds. When using the new query to return only the columns the visualization graph needs, it only takes around 20 seconds. 

cc @darla-leafnode @clarysabel

abatac:
Hi @darla-leafnode, I'm done addressing all the comments. Please check.

PR #716, opened 2019-12-10
IBP-3203 alias mapping

mderamos:
Kindly review. Thanks!

### DIFF
--- a/src/main/java/org/generationcp/middleware/operation/builder/StandardVariableBuilder.java
+++ b/src/main/java/org/generationcp/middleware/operation/builder/StandardVariableBuilder.java
@@ -79,6 +79,21 @@ public class StandardVariableBuilder extends Builder {
 		return standardVariables;
 	}
 
+	public List<StandardVariable> createStandardVariableWithAliasAsName(final List<Integer> standardVariableIds, final String programUUID) {
+		final List<StandardVariable> standardVariables = new ArrayList<>();
+		if (standardVariableIds != null && !standardVariableIds.isEmpty()) {
+			for (final Integer id : standardVariableIds) {
+				final Variable variable = this.getOntologyVariableDataManager().getVariable(programUUID, id, false);
+				final StandardVariable standardVariable = this.getStandardVariableTransformer().transformVariable(variable);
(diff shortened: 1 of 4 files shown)


src/main/java/org/generationcp/middleware/operation/builder/StandardVariableBuilder.java
```diff
@@ -79,6 +90,21 @@ public StandardVariable create(final int standardVariableId, final String progra
 		return standardVariables;
 	}
 
+	public List<StandardVariable> createStandardVariableWithAliasAsName(final List<Integer> standardVariableIds, final String programUUID) {
+		final List<StandardVariable> standardVariables = new ArrayList<>();
+		if (standardVariableIds != null && !standardVariableIds.isEmpty()) {
+			for (final Integer id : standardVariableIds) {
+				final Variable variable = this.getOntologyVariableDataManager().getVariable(programUUID, id, false);
```
abatac: You can just replace this code inside the loop with createStandardVariableWithAliasAsName(final int standardVariableId, final String programUUID);

src/main/java/org/generationcp/middleware/operation/builder/StandardVariableBuilder.java
```diff
@@ -69,6 +69,17 @@ public StandardVariable create(final int standardVariableId, final String progra
 		return this.getStandardVariableTransformer().transformVariable(variable);
 	}
 
+	public StandardVariable createStandardVariableWithAliasAsName(final int standardVariableId, final String programUUID) {
+
+		final Variable variable = this.getOntologyVariableDataManager().getVariable(programUUID, standardVariableId, false);
+
+		final StandardVariable standardVariable = this.getStandardVariableTransformer().transformVariable(variable);
+		if(variable.getAlias() != null && !variable.getName().isEmpty()) {
```
abatac: Maybe we can just have a separate field for Alias in StandardVariable, so that we preserve the actual ontology name.

abatac:
Hi @mderamos please check my comments. Thanks

PR #720, opened 2019-12-16
IBP-1971 Exclude Null Parent Gid

gelli-leafnode:
Exclude null parent id by checking if the option is present 
Catch null pointer exception to list with null parent


src/test/java/org/generationcp/middleware/service/pedigree/PedigreeServiceImplTest.java
```diff
@@ -106,7 +107,12 @@ public void testGetCrossExpansions() {
         final Germplasm generateRandomGermplasm = this.generateRandomGermplasmRecurringMaleParent();
         Set<Integer> set = new HashSet<>();
         set.add(generateRandomGermplasm.getGid());
-        Map<Integer, String> map = this.pedigreeService.getCrossExpansions(set, 5, this.crossExpansionProperties);
-        Assert.assertTrue(map.containsValue("-"));
+        try{
+            this.pedigreeService.getCrossExpansions(set, 5, this.crossExpansionProperties);
```
darla-leafnode: Put Assert.fail after this line:
`this.pedigreeService.getCrossExpansions(set, 5, this.crossExpansionProperties);`

src/test/java/org/generationcp/middleware/service/pedigree/AncestryTreeServiceTest.java
```diff
@@ -149,11 +149,57 @@ private Germplasm generateTestGermplasm(final int gid, final int methodId) {
 		germplasm.setGid(gid);
 		germplasm.setGnpgs(2);
 		final Method method = new Method(methodId);
+		if(methodId > 1) {
+			method.setMname("Backcross");
+			germplasm.setGpid1(0);
```
darla-leafnode: Maybe better to use non-zero as parent GIDs?

mderamos:
@gelli-leafnode  please resolve conflicts and check if the changes still works.

src/test/java/org/generationcp/middleware/service/pedigree/AncestryTreeServiceTest.java
gelli-leafnode: Removed this unit test. Unit test are in PedigreeServiceImplTest

PR #726, opened 2019-12-20
IBP-3294 Fix Performance of Import Data Discard

abatac:


src/main/java/org/generationcp/middleware/service/impl/dataset/DatasetServiceImpl.java
```diff
@@ -716,15 +710,19 @@ public void deleteDataset(final int datasetId) {
 	}
 
 	@Override
-	public void rejectDatasetDraftData(final Integer datasetId) {
+	public void rejectDatasetDraftData(final Integer studyId, final Integer datasetId) {
 		final List<Phenotype> phenotypes = this.daoFactory.getPhenotypeDAO().getDatasetDraftData(datasetId);
 		for (final Phenotype phenotype : phenotypes) {
 			if (StringUtils.isEmpty(phenotype.getValue())) {
+				// Set isChanged to true so that the derived traits that depend on it will be tagged as OUT_OF_SYNC later.
+				phenotype.setChanged(true);
```
nahuel-soldevilla: @abatac @darla-leafnode 
This method is rejecting draft data, right? why would we want to `reorganizePhenotypesStatus` here? 
I mean, until draft data is accepted, there shouldn't be any changes to phenotypes statuses.

darla-leafnode: Hi @nahuel-soldevilla - @abatac did not change any behavior here. Notice in master code that the call to update other variables as out of sync is already being as part of `deletePhenotype`. What he did was to make this out of sync update faster in reject and accepting draft data by doing it one-off instead as part of `deletePhenotype`.

I discussed with @christineguiang and she was able to verify that behavior is the same in master (ie. branch did not change behavior, just made the process faster). But we both believe you are correct that we shouldn't be updating phenotypes status until data is accepted. Hence, will include that fix in this branch as well.

darla-leafnode:
Hi @nahuel-soldevilla , @clarysabel - kindly review and merge.
Shiela did retesting of branch after the code to update phenotypes status was removed when rejecting draft data.

nahuel-soldevilla:
:+1:

PR #731, opened 2020-01-13
Upgrade hibernate version to 4.2.11.Final

clarysabel:
In this PR we are upgrading Hibernate version to `4.2.11.Final` 

This version was decided based on the limitation we have with our current sprint version (`4.1.6.RELEASE`). To be able to upgrade to Hibernate 5, we need to upgrade to `4.3.3.RELEASE` at minimum but considering most examples are based in `4.3.7.RELEASE`, I think we can just upgrade to that one. 

**We have faced different issues while trying this upgrade:** 
- In version 4, scalars can not be duplicated, so given the nature of our dynamic queries we had to include an **IF** condition before adding some scalars
- Generating experimental design did not complete after the upgrade. We were doing some research and after trying different approaches, we found that the generation table for `nd_experimentprop` worked well. It is still not clear to me why the combination of the generation types Table and Identity between `nd_experiment` and `nd_experiment_prop` worked well in Hibernate 3. It makes more sense that all tables under `nd_experiment` use Table as we are inserting in batch and Identity disables the batch option automatically
- In Hibernate 4, we need to flush the session in test environment before asserting. I have been researching if there is a way to automatically flush but seems the best practice is to programatically flush so we avoid false positives and negatives while testing. This is the reason why I have removed some `refresh()` and added some `flush()` 
- After many rounds of testing we have some performace issues in: **Saving imported cross list**, Saving advanced list, Create sub-observations being the first one the most significative. 
- We have agreed to try to upgrade our spring version as soon as possible to `4.3.7.RELEASE` so we can use Hibernate 5 

CC @darla-leafnode @nahuel-soldevilla


src/main/resources/liquibase/crop_changelog/17_1_0.xml
```diff
@@ -27,4 +27,20 @@
...
+        <preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="0">
+				select count(1) from sequence where sequence_name = 'nd_experimentprop'
+			</sqlCheck>
+        </preConditions>
+        <comment>
+            Create sequence entry for nd_experimentprop table generator
+        </comment>
+        <sql dbms="mysql" splitStatements="true">
+			insert into sequence(sequence_name, sequence_value)
```
darla-leafnode, line 41: Question, what was the 1) rationale and has been 2)  the effect of using sequence entry for table generation for `nd_experimentprop`?

PR #734, opened 2020-01-15
IBP-2909 - Import Crosses Optimization

mderamos:
Kindly Review. Thanks!


src/main/java/org/generationcp/middleware/pojos/germplasm/ImportedCrossParent.java
```diff
@@ -0,0 +1,47 @@
+package org.generationcp.middleware.pojos.germplasm;
+
+import org.pojomatic.annotations.AutoProperty;
+
+import java.io.Serializable;
+
+@AutoProperty
+public class ImportedCrossParent implements Serializable {
```
darla-leafnode: We have so many DTOs already in Middleware related to Germplasm. Let's reuse StudyGermplasm DTO here:
* designation
* germplasmId
* position = plotNo

src/main/java/org/generationcp/middleware/service/api/FieldbookService.java
```diff
@@ -984,4 +984,13 @@ Integer saveGermplasmList(List<Pair<Germplasm, GermplasmListData>> listDataItems
...
 	void saveWorkbookVariablesAndObservations(final Workbook workbook, final String programUUID);
+
+	/**
+	 * Retrieves the Plot Number to Imported Cross Parent Map
+	 *
+	 * @param studyId  - the Study Id
+	 * @param plotNos - the set of Plot Numbers
+	 * @return
+	 */
+	Map<Integer, ImportedCrossParent> getPlotNoToImportedCrossParentMap(Integer studyId, Set<Integer> plotNos);
```
darla-leafnode: When we use StudyGermplasmDTO, this would have to be renamed.

src/main/java/org/generationcp/middleware/dao/GermplasmDAO.java
```diff
@@ -1646,4 +1650,35 @@ public long countGermplasmByStudy(final Integer studyDbId) {
 			throw new MiddlewareQueryException(message, e);
 		}
 	}
+
+	public Map<Integer, StudyGermplasmDto> getPlotNoToStudyGermplasmDtoMap(final Integer studyId, final Set<Integer> plotNos) {
```
clarysabel: @darla-leafnode reviewing the code for the demo tomorrow, I dont get why we need to map this inside the DAO? can not this be done at the service level if getting the StudyGermplasmDto by studyId is useful for another feature?

darla-leafnode: Yes, actually the map can be actually be formed at the service-level.
@mderamos

PR #741, opened 2020-01-31
IBP-3283 Set Value of Experimental Design Variables 

gelli-leafnode:
Set Value for experimental design variables if not existing. 
Apply suggested code cleanup.


darla-leafnode:
@gelli-leafnode - I am rejecting this fix. 
While it is a workaround for imported studies, it makes the code more complicated than it should be. The proper fix is to change the way we import/save the experimental design parameter values during import.

PR #755, opened 2020-02-13
IBP-3448 AddedAttributeSort 

gelli-leafnode:
Add attribute addedColumnCurrentSort. 
If addedColumnCurrentSort is null return columnValuesMap keyset.

### DIFF
--- a/src/main/java/org/generationcp/middleware/domain/gms/GermplasmListNewColumnsInfo.java
+++ b/src/main/java/org/generationcp/middleware/domain/gms/GermplasmListNewColumnsInfo.java
@@ -2,6 +2,8 @@
 package org.generationcp.middleware.domain.gms;
 
 import java.io.Serializable;
+import java.lang.reflect.Array;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


src/main/java/org/generationcp/middleware/domain/gms/GermplasmListNewColumnsInfo.java
```diff
@@ -119,4 +122,20 @@ public void print(int indent) {
 		return new HashSet<>();
 	}
 
+	public List<String> getAddedColumnCurrentSort() {
```
darla-leafnode: As discussed, please work on a solution that will work for both cases when added columns have been saved and when re-opened again.

src/test/java/org/generationcp/middleware/data/initializer/GermplasmListNewColumnsInfoTestDataInitializer.java
```diff
@@ -16,11 +13,12 @@ public static GermplasmListNewColumnsInfo createGermplasmListNewColumnsInfo() {
 
 	public static GermplasmListNewColumnsInfo createGermplasmListNewColumnsInfo(final String columnHeader, final String columnValue) {
 		final GermplasmListNewColumnsInfo germplasmListNewColumnsInfo = new GermplasmListNewColumnsInfo(1);
-		final Map<String, List<ListDataColumnValues>> columnValuesMap = new HashMap<>();
+		final LinkedHashMap<String, List<ListDataColumnValues>> columnValuesMap = new LinkedHashMap<>();
```
darla-leafnode: Declare `columnValuesMap` as `Map`

PR #758, opened 2020-02-24
IBP-3379 Check For Existing Trial Instance on Import

gelli-leafnode:
Remove additional checking for identifying if trial instance already exists 
Create unit test

### DIFF
--- a/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
+++ b/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
@@ -1164,24 +1164,7 @@ public class DataImportServiceImpl extends Service implements DataImportService 
 				}
 			}
 		}
-		boolean hasDuplicateTrialInstances = false;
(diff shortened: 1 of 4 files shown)


src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -1164,24 +1164,7 @@ private void checkForExistingTrialInstance(final Workbook workbook, final Map<St
...
-
-			final int numberOfVariatesData = this.getPhenotypeDao()
-				.countVariatesDataOfStudy(
-					isMeansDataImport ? workbook.getMeansDatasetId() : workbook.getMeasurementDatesetId(),
-					variateIds);
-			if (numberOfVariatesData > 0) {
-				hasDuplicateTrialInstances = true;
-			}
-		}
-		if (hasDuplicateTrialInstances) {
```
darla-leafnode: As discussed, we need to handle still being able to import means dataset too for the same instance number.

PR #760, opened 2020-03-04
Ibp 3303 environments api architecture

clarysabel:
Creating PR on @darla-leafnode behalf


src/main/java/org/generationcp/middleware/dao/dms/EnvironmentDao.java
```diff
@@ -35,93 +28,61 @@
...
 import java.util.Map;
 import java.util.Set;
 
-/**
- * DAO class for {@link Geolocation}.
- */
-public class GeolocationDao extends GenericDAO<Geolocation, Integer> {
-
-	private static final Logger LOG = LoggerFactory.getLogger(GeolocationDao.class);
+public class EnvironmentDao extends GenericDAO<ExperimentModel, Integer> {
```
clarysabel: @darla-leafnode it means that we will have 2 different DAO's to refer to the same entity? if that's the case, then we should move the functions from ExperimentDao that refers to environments.

src/main/java/org/generationcp/middleware/dao/dms/EnvironmentDao.java
```diff
@@ -315,69 +277,60 @@ public TrialEnvironments getEnvironmentsForTraits(final List<Integer> traitIds,
...
-				+ "   and st.program_uuid = :programUUID" + "   and g.description = :locationDescription";
-			final Query query = this.getSession().createSQLQuery(sql);
-			query.setParameter("projectName", projectName);
-			query.setParameter("locationDescription", locationDescription);
-			query.setParameter("programUUID", programUUID);
-			final List<Integer> list = query.list();
-			if (list != null && !list.isEmpty()) {
-				return list.get(0);
-			}
+	public Integer getNextInstanceNumber(final Integer datasetId) {
```
clarysabel: @darla-leafnode we should convey in a unique term to refer to this and rename classes, functions, variables progressively to avoid confussions in the future (instance or environment)

src/main/java/org/generationcp/middleware/dao/dms/EnvironmentDao.java
```diff
@@ -315,69 +277,60 @@ public TrialEnvironments getEnvironmentsForTraits(final List<Integer> traitIds,
...
+		final StringBuilder sb = new StringBuilder();
+		sb.append("SELECT max(observation_unit_no) ");
+		sb.append(" FROM nd_experiment e ");
+		sb.append(" WHERE e.project_id = :datasetId and e.type_id = 1020" );
```
clarysabel: @darla-leafnode I have seen in many places that we are hardcoding some ids, can it be possible to use TermId instead?

src/main/java/org/generationcp/middleware/dao/dms/EnvironmentDao.java
```diff
@@ -417,20 +370,20 @@ public Integer getLocationIdByProjectNameAndDescriptionAndProgramUUID(
 		return tiMetadata;
 	}
 
-	public List<Geolocation> getEnvironmentGeolocationsForInstances(final Integer studyId, final List<Integer> instanceNumbers) {
-		List<Geolocation> returnList = new ArrayList<>();
+	public List<ExperimentModel> getEnvironmentsForInstances(final Integer studyId, final List<Integer> instanceNumbers) {
```
clarysabel: Same here @darla-leafnode 
should this be renamed to getEnvironmentsByEnvironmentNumber? 
Also this function should be moved to the new DAO. To me is very uncommon to have 2 DAO's for the same entity but I understand that it can get very long. So I dont have a strong position on this, but probably it will be easier to maintain merging both

PR #761, opened 2020-03-05
IBP-3523 Check If Method Is Present

gelli-leafnode:
If method is not present, return null as germplasm node. 
Add catch of nullpointer exception for proper handling of error.


darla-leafnode:
As discussed in our call, fix is not proper for real root cause for offending germplasm (`grplace != 0`).
Closing this PR, please make a new branch.

PR #763, opened 2020-03-06
IBP-3523 Add GRPLCE = 0 in Criteria

gelli-leafnode:
Add g.grplce = 0 in where clause of GermplasmDAO.getById

### DIFF
--- a/src/main/resources/liquibase/crop_changelog/14_2_0.xml
+++ b/src/main/resources/liquibase/crop_changelog/14_2_0.xml
@@ -1,0 +1,47 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="gelli" id="v14.2.0-1">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <sqlCheck expectedResult="1">
+                    SELECT CASE WHEN (SELECT COUNT(*) FROM germplsm WHERE methn = 0) > 0 THEN 1 ELSE 0 END;
+                </sqlCheck>
+                <sqlCheck expectedResult="2">
+                    SELECT COUNT(*) FROM methods WHERE mid IN (1,31);
+                </sqlCheck>
+            </and>
+        </preConditions>
+        <comment>Update records with germplasm.methn = 0 or germplasm.methn not existing to methods </comment>
+        <sql dbms="mysql" >
+            UPDATE germplsm SET methn =
+                CASE WHEN gnpgs &lt;= 0 AND methn = 0 THEN 31
+                     WHEN gnpgs &gt; 0 AND methn = 0 THEN 1
+                     WHEN gnpgs &gt; 0 AND methn &lt;&gt; 0 AND (SELECT COUNT(*) FROM methods WHERE mid = methn ) = 0 THEN 1
+                     WHEN gnpgs &lt;= 0 AND methn &lt;&gt; 0 AND (SELECT COUNT(*) FROM methods WHERE mid = methn ) = 0 THEN 31
+                     ELSE methn
+                END;
+        </sql>
+    </changeSet>
+    <changeSet id="v14.2.0-2" author="gelli">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <not>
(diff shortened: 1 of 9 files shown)


src/main/resources/liquibase/crop_changelog/14_2_0.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
```
darla-leafnode: Transfer to **14_3_0.xml** file and update changeset IDs as "v14.3.0-x"

src/main/resources/liquibase/crop_changelog/14_2_0.xml
```diff
@@ -0,0 +1,47 @@
...
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="gelli" id="v14.2.0-1">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <sqlCheck expectedResult="1">
+                    SELECT CASE WHEN (SELECT COUNT(*) FROM germplsm WHERE methn = 0) > 0 THEN 1 ELSE 0 END;
```
darla-leafnode: We should include in precondition whether there are any records with `methn` not existing in `methods` table.
The first precondition can be simplified/consolidated as:
```
 SELECT COUNT(*) > 0 FROM germplsm 
 WHERE methn = 0 OR NOT EXISTS (
       SELECT 1 FROM methods WHERE mid = methn);
```

src/main/resources/liquibase/crop_changelog/14_2_0.xml
```diff
@@ -0,0 +1,47 @@
...
+                </sqlCheck>
+                <sqlCheck expectedResult="2">
+                    SELECT COUNT(*) FROM methods WHERE mid IN (1,31);
+                </sqlCheck>
+            </and>
+        </preConditions>
+        <comment>Update records with germplasm.methn = 0 or germplasm.methn not existing to methods </comment>
+        <sql dbms="mysql" >
+            UPDATE germplsm SET methn =
```
darla-leafnode: For any changeset, we should only update the records we need to. Hence, we shouldn't be updating the whole `germplsm` table but only the records that we needed to (i.e. methn = 0 or methn doesn't exist in method table).

As such, we can simplify the script as:
UPDATE germplsm SET 
  methn = CASE WHEN gnpgs <= 0 THEN 31
	         ELSE 1
		END
WHERE methn = 0 OR NOT EXISTS (SELECT 1 FROM methods WHERE mid = methn);

`

src/main/resources/liquibase/crop_changelog/14_2_0.xml
```diff
@@ -0,0 +1,47 @@
...
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet author="gelli" id="v14.2.0-1">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <sqlCheck expectedResult="1">
+                    SELECT CASE WHEN (SELECT COUNT(*) FROM germplsm WHERE methn = 0) > 0 THEN 1 ELSE 0 END;
+                </sqlCheck>
+                <sqlCheck expectedResult="2">
+                    SELECT COUNT(*) FROM methods WHERE mid IN (1,31);
```
darla-leafnode: Just noting that I think it is safe to use method IDs 1 and 31 as it's included in common [methods](https://github.com/IntegratedBreedingPlatform/DBScripts/blob/master/merged/common/20_Data_Common_Methods.sql) seeded for all crops

src/main/resources/liquibase/crop_changelog/14_2_0.xml
```diff
@@ -0,0 +1,47 @@
...
+                END;
+        </sql>
+    </changeSet>
+    <changeSet id="v14.2.0-2" author="gelli">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <not>
+                    <foreignKeyConstraintExists foreignKeyName="fk_germplasm_methods_methn"></foreignKeyConstraintExists>
+                </not>
+                <sqlCheck expectedResult="0">
```
darla-leafnode: The 2 preconditions checking records in germplsm records can be consolidated together:
```
SELECT COUNT(*) > 0 FROM germplsm 
 WHERE methn = 0 OR NOT EXISTS (
       SELECT 1 FROM methods WHERE mid = methn);
```
Noticed I replaced the "NOT IN" with "NOT EXISTS" as that is more performant.

PR #779, opened 2020-04-06
IBP-3185 KSI Field Book - BrAPI Integration

abatac:


src/main/java/org/generationcp/middleware/dao/dms/PhenotypeDao.java
```diff
@@ -942,14 +943,17 @@ public Boolean containsAtLeast2CommonEntriesWithValues(final int projectId, fina
...
-				String x = (String) row[16]; // ROW
-				String y = (String) row[17]; // COL
+
+				String x = row[16] != null ? (String) row[16] : null; // ROW
+				String y = row[17] != null ? (String) row[17] : null; // COL
 				if (StringUtils.isBlank(x) || StringUtils.isBlank(y)) {
-					x = (String) row[11]; // fieldMapRow
-					y = (String) row[12]; // fieldMapCol
+					x = row[11] != null ? (String) row[11] : "1"; // fieldMapRow
+					y = row[12] != null ? (String) row[12] : "1"; // fieldMapCol
```
nahuel-soldevilla, line 951: `1, 1` for all :thinking:

src/main/java/org/generationcp/middleware/service/impl/dataset/DatasetServiceImpl.java
```diff
@@ -1018,6 +1023,14 @@ public void importDataset(final Integer datasetId, final Table<String, String, S
 							}
 
 						}
+						if (measurementVariable.getDataTypeId() == TermId.DATE_VARIABLE.getId()) {
+							try {
+								// In case the date is in yyyy-MM-dd format, try to parse it as number format yyyyMMdd
+								importedVariableValue = Util.convertDate(importedVariableValue, Util.FRONTEND_DATE_FORMAT, Util.DATE_AS_NUMBER_FORMAT);
+							} catch (ParseException e) {
+								// Do nothing
+							}
```
nahuel-soldevilla, line 1032: Can we use `tryParseDate`?

abatac: I just created a new method tryConvertDate to avoid the try-catch block here.

src/main/java/org/generationcp/middleware/dao/dms/PhenotypeDao.java
abatac: This also got me thinking. If we set these values to null, KSU fieldbook will fail to import a study, because the app expects these fields to have values, and it also uses these fields as secondary and tertiary IDs. Maybe we can change these to 0 as default?

nahuel-soldevilla: Mmm I see. I guess it's right to say that if there's no layout, then everything is in one "big" cell call `(1,1)`. `0` might work too but I don't know how Field book will take it. And `0` is problematic in some places. Yeah, let's keep `(1, 1)`

src/main/java/org/generationcp/middleware/service/impl/dataset/DatasetServiceImpl.java
nahuel-soldevilla: I guess that's fine. But another way could be to just ignore the old format and try to parse the new one.
And for isValid, another option would be to overload it with `isValid(date, format)`

@abatac

src/main/java/org/generationcp/middleware/dao/dms/PhenotypeDao.java
```diff
@@ -1010,11 +1014,11 @@ public Boolean containsAtLeast2CommonEntriesWithValues(final int projectId, fina
...
 				observation.setObservationVariableDbId(variableId);
 				observation.setObservationVariableName((String) result[3]);
-				observation.setObservationDbId((Integer) result[1]);
+				observation.setObservationDbId(String.valueOf((Integer) result[1]));
 				observation.setValue((String) result[4]);
-				observation.setObservationTimeStamp(Util.formatDateAsStringValue((Date) result[6], Util.FRONTEND_TIMESTAMP_FORMAT));
+				observation.setObservationTimeStamp((Date) result[6]);
 				// TODO
-				observation.setSeason(StringUtils.EMPTY);
+				observation.setSeason(null);
```
darla-leafnode, line 1021: We can just remove setting season to null, right?

PR #781, opened 2020-04-06
IBP-3613 Set default Seed Storage Location

clarysabel:


src/main/resources/liquibase/crop_changelog/14_4_0.xml
```diff
@@ -0,0 +1,100 @@
...
+					where l.ldefault = 1 and l.ltype = (select u.fldno from udflds u where u.ftable='LOCATION' and u.ftype='LTYPE' and fcode='SSTORE')
+				</sqlCheck>
+			</and>
+		</preConditions>
+
+		<comment>Set DSS as default storage location</comment>
+
+        <sql dbms="mysql">
+			update location set ldefault = 1 where
+			ltype = (select u.fldno from udflds u where u.ftable='LOCATION' and u.ftype='LTYPE' and fcode='SSTORE') and labbr='DSS'
```
nahuel-soldevilla, line 36: @clarysabel 
The `labbr` can be changed right? what if we get the first one with constraints `u.ftable='LOCATION' and u.ftype='LTYPE' and fcode='SSTORE'` and set that one as default?

src/main/resources/liquibase/crop_changelog/14_4_0.xml
```diff
@@ -0,0 +1,100 @@
...
+		</preConditions>
+		<comment>Set the default storage location to the lots with no locations</comment>
+		<sql dbms="mysql">
+			update ims_lot set locid = (select l.locid from location l where l.ltype = (select u.fldno from udflds u where u.ftable='LOCATION' and u.ftype='LTYPE' and fcode='SSTORE') and l.ldefault=1)
+			where locid is null;
+		</sql>
+	</changeSet>
+
+	<changeSet author="clarysabel" id="v14.4.0-4">
+		<preConditions onFail="MARK_RAN">
```
nahuel-soldevilla, line 57: maybe this should be `WARN`? in case the _set default_ phase had any problems. Same with the others?

src/main/resources/liquibase/crop_changelog/14_4_0.xml
clarysabel: will talk to Mariano about using locid = 6000 as our default, it can be changed in a near future when we have an interface for it

src/main/resources/liquibase/crop_changelog/14_4_0.xml
clarysabel: will do

nahuel-soldevilla: @clarysabel provided some info:
- `WARN` -> not possible, changeset is executed anyway
- `CONTINUE` -> won't keep a record of the skipped changeset other than the missing row

Best option for now is to mark them as `MARK_RAN`

PR #797, opened 2020-04-29
IBP-3679 Include Prior Versions Properties

gelli-leafnode:
Include all fields from prior versions.

### DIFF
--- a/src/main/java/org/generationcp/middleware/service/api/BrapiView.java
+++ b/src/main/java/org/generationcp/middleware/service/api/BrapiView.java
@@ -5,15 +5,15 @@ package org.generationcp.middleware.service.api;
  */
 public class BrapiView {
 
-	public static class BrapiV1_3 {
+	public interface BrapiV1_3 extends BrapiV1_2 {
 
 	}
 
-	public static class BrapiV1_2 {
+	public interface BrapiV1_2 {
 
 	}
 
-	public static class BrapiV2 {
+	public interface BrapiV2 extends BrapiV1_3 {
 
 	}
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/generationcp/middleware/service/api/BrapiView.java
```diff
@@ -5,15 +5,15 @@
...
 
-	public static class BrapiV1_3 {
+	public interface BrapiV1_3 extends BrapiV1_2 {
 
 	}
 
-	public static class BrapiV1_2 {
+	public interface BrapiV1_2 {
 
 	}
```
mderamos, line 15: Hi @gelli-leafnode , minor comment. Move this interface above the BrapiV1_3 interface for readability purposes.

src/main/java/org/generationcp/middleware/service/api/BrapiView.java
```diff
@@ -5,15 +5,15 @@
...
 
-	public static class BrapiV1_2 {
+	public interface BrapiV1_3 extends BrapiV1_2 {
 
 	}
 
-	public static class BrapiV2 {
+	public interface BrapiV2 extends BrapiV1_3 {
 
 	}
```
nahuel-soldevilla, line 18: @gelli-leafnode 

This workaround won't work for v2, right? I mean, for major releases the _deprecated_ fields should be removed, so if we add a field on v1.2 and deprecate it on v1.3, it should not be available on v2, but with this approach the field will be there. (e.g `PhenotypeSearchDTO.x`)
But of course, some fields **do** persist across major releases, a thing that this approach seems to solve.

cc @mderamos @darla-leafnode @abatac @clarysabel

gelli-leafnode: Hi @nahuel-soldevilla, 

Yes you are correct, deprecated fields will still be included for v2. I'll remove extension BrapiV1.3 from BrapiV2 and for the fields programId, treatments, trialDbId and trialName will have multiple versions (BrapiV1.3 and BrapiV2) (as suggested in IBP-3474) 

Thanks,
Gelli

PR #798, opened 2020-04-29
IBP-3441 Add FK to project.parent_project_id

gelli-leafnode:
Add liquibase script for adding fk to project.parent_project_id
IBP-3441-FKProjectParentId


src/main/resources/liquibase/crop_changelog/14_4_0.xml
```diff
@@ -104,5 +104,20 @@
 		<addForeignKeyConstraint baseTableName="ims_lot" baseColumnNames="locid" constraintName="location_fk_idx" referencedTableName="location"
 								 referencedColumnNames="locid"/>
 	</changeSet>
+	<changeSet author="gelli" id="v14.4.0-6">
```
darla-leafnode, line 107: We've released 14.4 so this changeset needs to be `14.4.1-x` in a `14_4_1.xml` file

PR #801, opened 2020-05-13
IBP-3749-AlterKeyPrefixDataType

gelli-leafnode:
Alter key_prefix data type from varchar(50) to varchar(255) matching the datatype of listdata.desig

### DIFF
--- a/src/main/resources/liquibase/crop_changelog/15_1_0.xml
+++ b/src/main/resources/liquibase/crop_changelog/15_1_0.xml
@@ -1,0 +1,25 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+				   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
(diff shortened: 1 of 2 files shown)


src/main/resources/liquibase/crop_changelog/15_1_0.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+				   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+	<changeSet author="gelli" id="v15.1.0-1">
+		<preConditions onFail="MARK_RAN">
+			<and>
+				<columnExists tableName="key_sequence_register" columnName="key_prefix"/>
```
darla-leafnode, line 10: I think the `<and>` condition is unnecessary because below check is redundant because we are already checking INFORMATION_SCHEMA.COLUMNS for such column:
`<columnExists tableName="key_sequence_register" columnName="key_prefix"/>`

PR #818, opened 2020-06-16
Ibp 3778 missing planting tests

clarysabel:


src/test/java/org/generationcp/middleware/service/impl/inventory/PlantingServiceImplIntegrationTest.java
```diff
@@ -0,0 +1,328 @@
...
+
+		try {
+			this.plantingService.savePlanting(userId, studyId, observationDatasetId, plantingRequestDto, TransactionStatus.CONFIRMED);
+		} catch (final MiddlewareRequestException e) {
+			assertThat(e.getErrorCode(), equalTo("planting.not.enough.inventory"));
+		}
+	}
+
+	@Test
+	public void test_SavePlanting_GroupTransactions_ThrowsException() {
```
nahuel-soldevilla, line 205: the last two don't throw an exception right?

clarysabel: Fixed names, also added assert by quantity

PR #819, opened 2020-06-16
IBP-3752 study inventory

nahuel-soldevilla:
- New study inventory search + filters
- Refactor LotDao/TransactionDao for SqlQueryParamBuilder


src/main/java/org/generationcp/middleware/dao/ims/TransactionDAO.java
```diff
@@ -753,4 +830,152 @@ public Transaction update(final Transaction transaction) {
 		this.getSession().flush();
 		return transaction;
 	}
+
+	public long countAllStudyTransactions(final Integer studyId, final StudyTransactionsRequest studyTransactionsRequest) {
```
clarysabel, line 834: param seems not to be used. Do we need to different functions? we usually assume that if the object to filter is null, then it returns or count everything without filtering.

PR #820, opened 2020-06-17
IBP-3402-DataSetColumns

gelli-leafnode:
Siblings variables includes the default columns to be displayed in view summary
Created list for default columns to be included to filtered variables


mderamos:
Discussed a better solution with @gelli-leafnode

PR #821, opened 2020-06-18
IBP-3773 Replace study entry

darla-leafnode:


src/main/java/org/generationcp/middleware/dao/SampleDao.java
```diff
@@ -384,6 +392,22 @@ public boolean hasSamples(final Integer studyId) {
...
+		final List queryResults;
+		try {
+			final SQLQuery query = this.getSession().createSQLQuery(SQL_STUDY_ENTRY_HAS_SAMPLES);
+			query.setParameter("studyId", studyId);
+			query.setParameter("entryId", entryId);
+			queryResults = query.list();
+
+		} catch (final HibernateException he) {
+			throw new MiddlewareException(
+					"Unexpected error in executing hasSamples(studyId = " + studyId + ") query: " + he.getMessage(),
```
abatac, line 405: Fix the error message, hasSamples should be studyEntryHasSamples.

src/main/java/org/generationcp/middleware/service/impl/study/StudyGermplasmServiceImpl.java
```diff
@@ -101,6 +104,54 @@ public long countStudyGermplasmByEntryTypeIds(final int studyId, final List<Stri
...
+
+	@Override
+	public StudyGermplasmDto replaceStudyGermplasm(final int studyId, final int entryId, final int gid, final String crossExpansion) {
+		// Check first that new GID is valid
+		final Germplasm newGermplasm = this.daoFactory.getGermplasmDao().getById(gid);
+		if (newGermplasm == null) {
+			throw new MiddlewareException("Invalid GID: " + gid);
+		}
+		final StockDao stockDao = this.daoFactory.getStockDao();
+		// Copy from old entry: entry type, entry #, entry code
```
abatac, line 127: Do we still need to copy the seed source from old entry?

src/main/java/org/generationcp/middleware/dao/SampleDao.java
darla-leafnode: Done

src/main/java/org/generationcp/middleware/service/impl/study/StudyGermplasmServiceImpl.java
darla-leafnode: The germplasm to be used is not coming for a germplasm list so there is no seed source to copy from. I don't think it's correct to retain the old seed source value, if any, as it's for a different germplasm

PR #823, opened 2020-06-26
IBP-3402-DataSetColumns

gelli-leafnode:
Remove previous fix
Add new condition for filtering variables, exclude variables with variable type Environment and Study Details 
Modify Unit test

### DIFF
--- a/src/main/java/org/generationcp/middleware/operation/builder/DataSetBuilder.java
+++ b/src/main/java/org/generationcp/middleware/operation/builder/DataSetBuilder.java
@@ -14,7 +14,6 @@ package org.generationcp.middleware.operation.builder;
 import com.jamonapi.Monitor;
 import com.jamonapi.MonitorFactory;
 import org.apache.commons.lang3.StringUtils;
-import org.generationcp.middleware.dao.dms.DmsProjectDao;
 import org.generationcp.middleware.domain.dms.DMSVariableType;
 import org.generationcp.middleware.domain.dms.DataSet;
 import org.generationcp.middleware.domain.dms.DatasetReference;
@@ -34,7 +33,6 @@ import org.generationcp.middleware.pojos.dms.DmsProject;
 import org.generationcp.middleware.util.StringUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
 
 import javax.annotation.Resource;
 import java.util.ArrayList;
@@ -194,10 +192,6 @@ public class DataSetBuilder extends Builder {
 			}
 		}
 		return newVariables;
-	}
-
-	public List<Integer> getVariablesOfSiblingDatasets(final int datasetId) {
-		return this.getProjectPropertyDao().getVariablesOfSiblingDatasets(datasetId);
 	}
 
 	public VariableTypeList filterVariables(final VariableTypeList variables) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/generationcp/middleware/operation/builder/DataSetBuilder.java
```diff
@@ -209,7 +206,8 @@ public VariableTypeList filterVariables(final VariableTypeList variables, final
 		final VariableTypeList newList = new VariableTypeList();
 		if (variables != null && !variables.getVariableTypes().isEmpty()) {
 			for (final DMSVariableType variable : variables.getVariableTypes()) {
-				if (!filters.contains(variable.getId()) || this.DEFAULT_DATASET_COLUMNS.contains(variable.getId())) {
+				if (!filters.contains(variable.getId()) || variable.getId() == TermId.TRIAL_INSTANCE_FACTOR.getId()
+				|| (!variable.getVariableType().equals(VariableType.ENVIRONMENT_DETAIL) && !variable.getVariableType().equals(VariableType.STUDY_DETAIL))) {
```
mderamos, line 210: Hi @gelli-leafnode. Since we're already filtering out ENVIRONMENT and STUDY DETAILS, do we still need the filters list?
I think the variables shown in the table are VariableType.GERMPLASM_DESCRIPTOR, VariableType.TRAIT and TRIAL_INSTANCE_FACTOR. Can you double check and just filter those? Also check the scenario with Treatment Factors.

src/main/java/org/generationcp/middleware/operation/builder/DataSetBuilder.java
```diff
@@ -24,6 +24,7 @@
 import org.generationcp.middleware.domain.etl.MeasurementVariable;
 import org.generationcp.middleware.domain.etl.Workbook;
 import org.generationcp.middleware.domain.oms.TermId;
+import org.generationcp.middleware.domain.ontology.VariableType;
 import org.generationcp.middleware.enumeration.DatasetTypeEnum;
 import org.generationcp.middleware.exceptions.MiddlewareQueryException;
 import org.generationcp.middleware.helper.VariableInfo;
```
abatac, line 30: Remove unused imports

src/main/java/org/generationcp/middleware/operation/builder/DataSetBuilder.java
```diff
@@ -197,15 +197,15 @@ protected VariableTypeList filterDatasetVariables(final VariableTypeList variabl
 		return newVariables;
 	}
 
-	private List<Integer> getVariablesOfSiblingDatasets(final int datasetId) {
+	public List<Integer> getVariablesOfSiblingDatasets(final int datasetId) {
```
abatac, line 200: This is not used anymore, do you know the reason why this method is used before? Have you checked other functionalities that call this method?

abatac: getVariablesOfSiblingDatasets was refactored last year, maybe that caused the bug? 

https://github.com/IntegratedBreedingPlatform/Middleware/commit/6645558deb70e30f4752f94dd98a148fa5d53e1d#diff-ae8f7c46d0cbcc2cb851c3bd8d7f09e0L162

gelli-leafnode: Hi Aldrin, upon investigation before, this method is used to exclude variables that exists both in study's dataset. I believe its purpose is to exclude DATASET types. Comparing to Browse Study, w/c filters only by variable type. It is only used in buildCompleteDataset.isMeasurementDataset

src/main/java/org/generationcp/middleware/operation/builder/DataSetBuilder.java
gelli-leafnode: Remove sibling variables from filters.

PR #825, opened 2020-06-30
IBP-3557 Removed saving or updating of environment variables

mderamos:


darla-leafnode:
As discussed, the addition of **excludeEnvironmentVariables** flag complicates the WorkbookSaver class even more. Let's remove the minimal changes confirmed to be unused by the Manage Studies workflow.

PR #827, opened 2020-07-06
IBP-3750 germplasm manager - create inventory lots

nahuel-soldevilla:


src/main/java/org/generationcp/middleware/domain/inventory/manager/LotItemDto.java
```diff
@@ -9,8 +9,14 @@
 
 	private Integer gid;
 
+	private Integer storageLocationId;
+
+	/** Ignored if storageLocationId is present */
 	private String storageLocationAbbr;
 
+	private Integer scaleId;
```
clarysabel, line 17: rename to unitId

src/main/java/org/generationcp/middleware/domain/inventory/manager/LotItemDto.java
```diff
@@ -9,8 +9,14 @@
...
+
+	/** Ignored if scaleId is present */
```
clarysabel, line 19: This one should be unitName instead to be consistent with the naming we decided to use, but I think we will have to change it in Inventory Manager. We can tackle this one later if it adds extra work

src/main/java/org/generationcp/middleware/service/impl/inventory/LotServiceImpl.java
```diff
@@ -135,47 +136,68 @@ public void updateLots(final List<ExtendedLotDto> lotDtos, final LotUpdateReques
 	}
 
 	@Override
-	public void saveLotsWithInitialTransaction(final CropType cropType, final Integer userId, final List<LotItemDto> lotItemDtos) {
+	public List<String> saveLots(final CropType cropType, final Integer userId, final List<LotItemDto> lotItemDtos) {
```
clarysabel, line 139: I think the previous name was more appropiated because we are saving lots and opening a deposit transaction

src/main/java/org/generationcp/middleware/domain/inventory/manager/LotItemDto.java
nahuel-soldevilla: I'll include later with the unitName refactor

src/main/java/org/generationcp/middleware/service/impl/inventory/LotServiceImpl.java
nahuel-soldevilla: now the initial transaction is optional

PR #852, opened 2020-08-07
IBP-3916 Refactor Graphical-Filtering to Graphical-Queries

nahuel-soldevilla:
Liquibase
- create new menu inside information management, new
permissions and **assign those to every role that previously had
BROWSE_STUDIES permission**

@clarysabel @cuenyad @mcrimi


src/main/resources/liquibase/workbench_changelog/15_4_0.xml
```diff
@@ -81,4 +81,38 @@
...
+            set @sidebar_category_link_id = (select sidebar_category_link_id from workbench_sidebar_category_link where tool_name = 'graphical_queries');
+
+            insert into permission(name, description, parent_id, workbench_sidebar_category_link_id)
+            values ('GRAPHICAL_QUERIES', 'Graphical Queries', @information_management_id, @sidebar_category_link_id);
+            set @last_permission_id = (select last_insert_id());
+            insert into role_type_permission(role_type_id, permission_id, selectable) values (1, @last_permission_id, 1);
+            insert into role_type_permission(role_type_id, permission_id, selectable) values (2, @last_permission_id, 1);
+            insert into role_type_permission(role_type_id, permission_id, selectable) values (3, @last_permission_id, 1);
+
+            # associate new 'GRAPHICAL_QUERIES' permission to every role that has 'BROWSE_STUDIES'
```
clarysabel, line 109: Is it needed when the role already contains INFORMATION_MANAGEMENT permission? Seems like in that case we should not inset a new role_permission  @nahuel-soldevilla

nahuel-soldevilla: @clarysabel good point, but I think in this case it's selecting only the roles that specifically has BROWSE_STUDIES, so if any role has INFORMATION_MANAGEMENT it probably won't have BROWSE_STUDIES (unless some manual db manipulation has taken place)
Right? let me know if I'm missing something

clarysabel: Rigth, they should not have both BROWSE_STUDIES and INFORMATION_MANAGEMENT unless someone manually changed the data

PR #855, opened 2020-08-11
IBP-3927 Added foreign key to germplasm_study_source gid

mderamos:


src/main/resources/liquibase/crop_changelog/15_4_0.xml
```diff
@@ -166,4 +166,26 @@
...
+		<preConditions onFail="MARK_RAN">
+            <and>
+                <not>
+                    <foreignKeyConstraintExists foreignKeyName="germplasm_study_source_gid_fk"></foreignKeyConstraintExists>
+                </not>
+            </and>
+		</preConditions>
+		<comment>Delete germplasm_study_source entries with no germplsm</comment>
+		<sql dbms="mysql" splitStatements="true">
+			CREATE TEMPORARY TABLE non_existent_gids
```
darla-leafnode, line 180: I don't see the need to create a temporary table. 
Can't we just a single script like:
```
DELETE FROM germplasm_study_source 
WHERE NOT EXISTS (SELECT 1 FROM germplsm g WHERE g.gid = germplasm_study_source.gid);
```

src/main/resources/liquibase/crop_changelog/15_4_0.xml
```diff
@@ -166,4 +166,21 @@
 			ALTER TABLE search_request CHANGE COLUMN parameters parameters MEDIUMTEXT NOT NULL ;
 		</sql>
 	</changeSet>
+
+	<changeSet author="mderamos" id="v15.4.0-7">
+		<preConditions onFail="MARK_RAN">
+            <and>
```
darla-leafnode, line 172: I don't think we need `<and>` clause here as there's only one precondition?

PR #860, opened 2020-08-17
IBP-2770-UseFieldParentValueAsFieldMapLocation

gelli-leafnode:
Remove override process for location if location is not null || eq. to 1
Modified query in getAllFieldMapsInBlockByTrialInstanceId - return fieldmap parent as sitelocationid 
Fix failing unit test

### DIFF
--- a/src/main/java/org/generationcp/middleware/dao/dms/ExperimentPropertyDao.java
+++ b/src/main/java/org/generationcp/middleware/dao/dms/ExperimentPropertyDao.java
@@ -26,6 +26,7 @@ import org.generationcp.middleware.domain.fieldbook.FieldMapTrialInstanceInfo;
 import org.generationcp.middleware.domain.oms.TermId;
 import org.generationcp.middleware.exceptions.MiddlewareQueryException;
 import org.generationcp.middleware.manager.Season;
+import org.generationcp.middleware.pojos.LocdesType;
 import org.generationcp.middleware.pojos.dms.ExperimentProperty;
 import org.generationcp.middleware.util.Debug;
 import org.hibernate.Criteria;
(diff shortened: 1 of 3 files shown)


src/main/java/org/generationcp/middleware/dao/dms/ExperimentPropertyDao.java
```diff
@@ -391,8 +391,8 @@
 				trial.setInstanceId((Integer) row[3]);
 				trial.setSiteName((String) row[4]);
 				trial.setLocationName((String) row[4]);
-				if (row[5] != null && NumberUtils.isNumber((String) row[5])) {
-					trial.setLocationId(Integer.valueOf((String) row[5]));
+				if (row[5] != null && NumberUtils.isNumber(row[5].toString())) {
```
abatac, line 394: Hi @gelli-leafnode , Use String.valueOf to prevent NPE in toString() when object is null

src/main/java/org/generationcp/middleware/operation/saver/GeolocationPropertySaver.java
```diff
@@ -26,9 +26,6 @@ public void saveFieldmapProperties(final List<FieldMapInfo> infos) throws Middle
 					int locationId = trial.getInstanceId();
 					if (trial.getInstanceId() != null && trial.getInstanceId().intValue() == 1) {
 						locationId = this.getExperimentModelSaver().moveStudyToNewGeolocation(info.getFieldbookId());
-					}
```
abatac, line 29: Why did you remove null checking for trial getLocationId?

abatac (changes requested):

src/main/java/org/generationcp/middleware/operation/saver/GeolocationPropertySaver.java
gelli-leafnode: Hi @abatac, will revert checking if location is not null before saving

abatac (approved):

PR #864, opened 2020-08-20
IBP-3961 search transactions refactor

cuenyad:
Included:
https://github.com/IntegratedBreedingPlatform/BMSAPI/pull/389

Please, Review.

### DIFF
--- a/src/main/java/org/generationcp/middleware/dao/ims/TransactionDAO.java
+++ b/src/main/java/org/generationcp/middleware/dao/ims/TransactionDAO.java
@@ -254,9 +254,9 @@ public class TransactionDAO extends GenericDAO<Transaction, Integer> {
 		+ "' WHEN trntype = " + TransactionType.ADJUSTMENT.getId() + "  THEN '" + TransactionType.ADJUSTMENT.getValue()
 		+ "' END) AS transactionType,"//
 		+ "    tr.trnqty AS amount,"//
-		+ "(SELECT SUM(CASE WHEN transaction.trnstat = " + TransactionStatus.CONFIRMED.getIntValue()
+		+ " IF (:lotAggregatedData, (SELECT SUM(CASE WHEN transaction.trnstat = " + TransactionStatus.CONFIRMED.getIntValue()
 		+ " OR (transaction.trnstat = " + TransactionStatus.PENDING.getIntValue() + " AND transaction.trntype = " + TransactionType.WITHDRAWAL.getId() + ") "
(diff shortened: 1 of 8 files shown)


src/main/java/org/generationcp/middleware/dao/ims/TransactionDAO.java
```diff
@@ -254,9 +254,9 @@ private void createTransactionRow(final List<TransactionReportRow> transactionRe
 		+ "' WHEN trntype = " + TransactionType.ADJUSTMENT.getId() + "  THEN '" + TransactionType.ADJUSTMENT.getValue()
 		+ "' END) AS transactionType,"//
 		+ "    tr.trnqty AS amount,"//
-		+ "(SELECT SUM(CASE WHEN transaction.trnstat = " + TransactionStatus.CONFIRMED.getIntValue()
+		+ " IF (:lotAggregatedData, (SELECT SUM(CASE WHEN transaction.trnstat = " + TransactionStatus.CONFIRMED.getIntValue()
 		+ " OR (transaction.trnstat = " + TransactionStatus.PENDING.getIntValue() + " AND transaction.trntype = " + TransactionType.WITHDRAWAL.getId() + ") "
-		+ "  THEN transaction.trnqty ELSE 0 END) FROM ims_transaction transaction WHERE  transaction.lotid = lot.lotid ) AS availableBalance, " //
+		+ "  THEN transaction.trnqty ELSE 0 END) FROM ims_transaction transaction WHERE  transaction.lotid = lot.lotid ), null) AS availableBalance, " //
```
nahuel-soldevilla, line 259: I tested this on centos-1 / maize (`ims_transactions` count = 110706; `ims_lots` count = 107393)
and I'm not seeing any noticeable difference between executions.
I've tested:

Case 1: No `availableBalance` sub-query (just removing the `(select ...  ) AS availableBalance,`)
Case 2:  lotAggregatedData=false -> `IF (0, (SELECT `
Case 3:  lotAggregatedData=true -> `IF (1, (SELECT `

centos-1-available-balance-performance

I'm not sure exactly why that is, maybe because we are already joining with lot :thinking: 
I also looked at the query plan, but I saw no difference, although I'm not sure if its relevant in the case of sub-queries.

**But in conclusion, my opinion is that this optimization is not worth it.**

cc @clarysabel @cuenyad

clarysabel:
Closing this PR, after some research we noticed that the performance did not suffer any degradation

PR #866, opened 2020-08-24
IBP-3960 Return notes property even if it is null

abatac:


abatac:
> Hey @abatac
> I think there's a better way to solve this [warning from datatables](https://datatables.net/manual/tech-notes/4), we can use `defaultContent` and set a default value when the column is missing or null.
> And the good thing is that we can set it in columnDefs and apply it for all columns, which is specially useful because we usually have null columns (or even omitted fields)
> 
> ```js
> // inventory-tab.js
> $scope.dtColumnDefs =  [
> ...
> 		{
> 			targets: "_all",
> 			orderable: false,
> 			defaultContent: "-"
> 		}
> ```

Yup this is a good solution. Will implement this.

PR #872, opened 2020-08-25
IBP-3793 Retrieve property from system environment variables

grg021:
#IBP-3793

### DIFF
--- a/src/main/java/org/generationcp/middleware/hibernate/DataSourceProperties.java
+++ b/src/main/java/org/generationcp/middleware/hibernate/DataSourceProperties.java
@@ -73,10 +73,10 @@ public class DataSourceProperties {
 
 	public DataSourceProperties(final Properties properties) {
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/generationcp/middleware/hibernate/DataSourceProperties.java
```diff
@@ -92,6 +92,14 @@ public DataSourceProperties(final Properties properties) {
 
 	}
 
+	private String getPropertyValue(final String envVariable, final Properties properties, final String defaultValue, final String property) {
+		final String valueInEnvironment = System.getenv(envVariable);
+		if (Strings.isNullOrEmpty(valueInEnvironment)) {
```
darla-leafnode, line 97: Can be simplified as:
`return Strings.isNullOrEmpty(valueInEnvironment)? getPropertyValue(properties, defaultValue, property) : valueInEnvironment;`

darla-leafnode (approved):

PR #874, opened 2020-08-26
Rename Variable Type: Study Condition to Environment Condition. Renam…

clarysabel:
…e enum

IBP-3969


clarysabel:
I created https://ibplatform.atlassian.net/browse/IBP-3981 to solve the issue and left a comment in the ticket.

PR #876, opened 2020-08-27
IBP-3732 Fill blank GROUPGID in observation and suboservation tables

abatac:


src/main/java/org/generationcp/middleware/service/impl/study/StudyGermplasmServiceImpl.java
```diff
@@ -54,7 +59,8 @@ public StudyGermplasmServiceImpl(final HibernateSessionProvider sessionProvider)
 			studyGermplasmDto.setSeedSource(this.findStockPropValue(TermId.SEED_SOURCE.getId(), stockModel.getProperties()));
 			studyGermplasmDto.setCheckType(Integer.valueOf(this.findStockPropValue(TermId.ENTRY_TYPE.getId(), stockModel.getProperties())));
 			studyGermplasmDto.setStockIds(stockIdsMap.getOrDefault(stockModel.getGermplasm().getGid(), ""));
-			studyGermplasmDto.setGroupId(stockModel.getGermplasm().getMgid());
+			final String groupGid = this.findStockPropValue(TermId.GROUPGID.getId(), stockModel.getProperties());
+			studyGermplasmDto.setGroupId(groupGid != null ? Integer.valueOf(groupGid) : 0);
```
darla-leafnode, line 63: Is it correct to set the groupGID value to 0 if the property is not present?

abatac: I think so, because the default value of mgid is 0.

src/main/resources/liquibase/crop_changelog/15_4_1.xml
```diff
@@ -173,6 +173,37 @@
 		</sql>
 	</changeSet>
 
+	<changeSet id="v15.4.1-10" author="aldrin">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<sqlCheck expectedResult="0">
+					-- Check if there are stocks that don't have GROUPGID stock property.
+					SELECT COUNT(*) FROM stock s
```
darla-leafnode, line 181: The precondition query can be improved by:
1. not performing a join to `germplsm` table
2. doing a `NOT EXISTS `instead of `LEFT JOIN` to `stockprop` table
            `AND NOT EXISTS (select 1 from stockprop sp WHERE sp.stock_id = s.stock_id AND sp.type_id = 8330);`

src/main/resources/liquibase/crop_changelog/15_4_1.xml
```diff
@@ -173,6 +173,37 @@
...
+		<sql dbms="mysql">
+			-- Value of GROUPGID should be the mgid of the germplasm.
+			-- Only create GROUPGID stockprop for studies with GROUPGID variable in projectprop.
+			INSERT INTO stockprop (`stock_id`, `type_id`, `value`, `rank`)
+			SELECT s.stock_id, 8330, g.mgid, IFNULL(MAX(spRank.rank), 0) + 1
+			FROM stock s
+			INNER JOIN germplsm g ON s.dbxref_id = g.gid
+            INNER JOIN project plotDataset ON plotDataset.parent_project_id = s.project_id AND plotDataset.dataset_type_id = 4
+            INNER JOIN projectprop ppPlot ON ppPlot.project_id = plotDataset.project_id AND ppPlot.variable_id = 8330
+			LEFT JOIN stockprop sp ON sp.stock_id = s.stock_id AND sp.type_id = 8330
```
darla-leafnode, line 200: Same here do `NOT EXISTS` instead of `LEFT JOIN to stockprop sp`

darla-leafnode (approved):

PR #882, opened 2020-09-03
IBP-4034 Fix error in executing calculated variable.

abatac:
IBP-4034


src/main/java/org/generationcp/middleware/service/impl/dataset/DatasetServiceImpl.java
```diff
@@ -510,7 +510,7 @@ void resolveObservationStatus(final Integer variableId, final Phenotype phenotyp
 	@Override
 	public void updateDependentPhenotypesAsOutOfSync(final Integer variableId, final Set<Integer> observationUnitIds) {
 		final List<Formula> formulaList = this.daoFactory.getFormulaDAO().getByInputId(variableId);
-		if (!formulaList.isEmpty()) {
+		if (!formulaList.isEmpty() && !observationUnitIds.isEmpty()) {
```
darla-leafnode, line 513: Can we use `CollectionUtils.isEmpty` for both checking `formulaList` and `observationunitIds` parameter?

darla-leafnode (approved):

PR #883, opened 2020-09-07
IBP-4010-VariableAliasForStudySettingsGermplasmDescriptor

gelli-leafnode:
Modify query for findPlotDatasetVariablesByTypesForStudy if variable type is Germplasm Descriptor (use alias if available) 
For getObservationUnitTableQuery.germplasmDescriptor use cvtermId if available to get value
For setting variable if alias is available use as key


src/main/java/org/generationcp/middleware/service/impl/dataset/DatasetServiceImpl.java
```diff
@@ -1177,8 +1177,14 @@ private static boolean isDateOrCharacterDataType(final MeasurementVariable measu
 	void addStudyVariablesToUnitRows(final List<ObservationUnitRow> observationUnits, final List<MeasurementVariable> studyVariables) {
 		for (final ObservationUnitRow observationUnitRow : observationUnits) {
 			for (final MeasurementVariable measurementVariable : studyVariables) {
+				final String key;
+				if (measurementVariable.getAlias() != null && !measurementVariable.getAlias().equals("")) {
```
abatac, line 1181: You can use StringUtils.isEmpty.

src/main/java/org/generationcp/middleware/service/impl/dataset/DatasetServiceImpl.java
```diff
@@ -1177,8 +1177,14 @@ private static boolean isDateOrCharacterDataType(final MeasurementVariable measu
...
+					key = measurementVariable.getAlias();
+				} else {
+					key = measurementVariable.getName();
+				}
 				observationUnitRow.getVariables()
-					.put(measurementVariable.getName(), new ObservationUnitData(measurementVariable.getValue()));
+					.put(key, new ObservationUnitData(measurementVariable.getValue()));
```
abatac, line 1187: Is there any effect on saving/retrieving/export of data if the germplasm descriptor has an alias?

gelli-leafnode: Hi @abatac , 
Upon checking saving/export/retrieve result is the same as master.

src/test/java/org/generationcp/middleware/service/impl/dataset/DatasetServiceImplTest.java
```diff
@@ -420,6 +420,18 @@ public void testaddStudyVariablesToUnitRows() {
 		Assert.assertNotNull(observationUnitRow.getVariables().get(trialInstanceVariable.getName()));
 	}
 
+	@Test
+	public void testaddStudyVariablesAliasToUnitRows() {
```
clarysabel, line 424: @gelli-leafnode please use camelCase

PR #884, opened 2020-09-07
IBP-4030-ImsLotStockId-Collation

gelli-leafnode:
Modify ims_lot.stock_id collation from utf8_general_ci to utf8_bin to support case sensitive


gelli-leafnode:
Closing this PR, Change no longer needed, stock id should be case insensitive

PR #888, opened 2020-09-16
IBP-3937 add germplasm UUID

mderamos:
Kindly review. Thanks!


src/main/resources/liquibase/crop_changelog/16_1_0.xml
```diff
@@ -0,0 +1,166 @@
...
+			IF useUuid = 0	THEN
+
+			WHILE(hasNullUuidCount > 0) DO
+
+			UPDATE germplsm g
+			SET germplsm_uuid = CONCAT((SELECT plot_code_prefix
+			FROM workbench.workbench_crop wc
+			WHERE db_name = DATABASE()),
+			'G',
+			RIGHT(sha1(CONCAT(uuid(), g.gid)), 15))
```
abatac, line 77: In Java code, the suffix is based on RandomUtils.randomAlphanumeric not sha1. Is there any reason for that?

src/main/resources/liquibase/crop_changelog/16_1_0.xml
```diff
@@ -0,0 +1,166 @@
...
+			WHERE g.germplsm_uuid IS NULL;
```
abatac, line 78: Is there other way to uniquely generate non uuid value?

src/main/resources/liquibase/crop_changelog/16_1_0.xml
```diff
@@ -0,0 +1,166 @@
...
+			<column name="germplsm_uuid" type="VARCHAR(36)">
+				<constraints nullable="true"/>
+			</column>
+		</addColumn>
+	</changeSet>
+
+	<changeSet author="mderamos" id="v16.1.0-2">
+		<preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="1">
+				SELECT
```
abatac, line 26: Why does it need to drop this procedure when it's not yet created?

mderamos: It's for safety purposes. I'm not sure if it's the standard but I've seen multiple liquibase scripts doing the same, dropping a procedure before creating it.

src/main/resources/liquibase/crop_changelog/16_1_0.xml
mderamos: I don't think there is. It's also what the BA team suggested. The ticket was tested in both centos-2 and 4 servers. It took 23:11minutes to complete the execution of liquibase scripts for use_uuid=0 for ALL crops.

src/main/resources/liquibase/crop_changelog/16_1_0.xml
mderamos: It follows the same logic in generating the lotuuid. Since uuid() is a MySQL method, RandomUtils.randomAlphanumeric is used to somewhat replicate the generation of unique string using the said method.

src/main/resources/liquibase/crop_changelog/16_1_0.xml
nahuel-soldevilla: To give more context on how this approach started: The [v13.3.0-4 changeset](https://github.com/IntegratedBreedingPlatform/Middleware/blob/e61b9aae8719384033e1655eb397642bd41fe1e9/src/main/resources/liquibase/crop_changelog/13_3_0.xml#L61), is in turn a simplified version of the original plot_id (obs_unit_id) migration: 
https://github.com/IntegratedBreedingPlatform/Middleware/commit/70b4eb14111602f5fd511cc4327be35a51038db8, an approach that also checks for duplicates in several iterations.

The first thing that I tried with plot_id was mysql `rand()`. I tested it on cimmyt db with an nd_experiment table of ~ 22 million rows, but I saw that the approach generated a lot of duplicates. The update was something like this:
```sql
DECLARE chars CHAR(62)  DEFAULT 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789';

UPDATE test_table
SET plot_id = CONCAT((SELECT plot_code_prefix
  FROM
	workbench.workbench_crop wc
  WHERE
	db_name = DATABASE()),
 'P', concat(
	 substring(chars, rand() * 62 + 1, 1), 
	 substring(chars, rand() * 62 + 1, 1),
	 substring(chars, rand() * 62 + 1, 1),
	 substring(chars, rand() * 62 + 1, 1),
	 substring(chars, rand() * 62 + 1, 1),
	 substring(chars, rand() * 62 + 1, 1)))
WHERE plot_id IS NULL;
```
After each iteration I had to remove duplicates, but I couldn't measure how long it'll take, I stopped testing after several hours.

Then I decided to try something involving `uuid()`, which is the approach wee see in https://github.com/IntegratedBreedingPlatform/Middleware/commit/70b4eb14111602f5fd511cc4327be35a51038db8. 
The `sha1()` part was so that the final result look more like a random string, given that [myql's uuid v1](https://dev.mysql.com/doc/refman/5.6/en/miscellaneous-functions.html#function_uuid) is a timestamp-based uuid.
I don't know why, but on `nd_experiment`  with 22M rows, even this alternative generated some duplicates. so I had to run the statement several times, checking for duplicates after each iteration. 

Luckily in the case of `ims_lot` and `germplsm` we don't see these "duplicates", and so the script is more simple and the update can be done in one pass.

cc @mderamos @abatac @darla-leafnode @clarysabel

src/main/java/org/generationcp/middleware/service/FieldbookServiceImpl.java
```diff
@@ -340,7 +340,7 @@ public Integer saveNurseryAdvanceGermplasmList(final List<Pair<Germplasm, List<N
 					if (germplasm.getLgid() == null) {
 						germplasm.setLgid(germplasm.getGid() != null ? germplasm.getGid() : Integer.valueOf(0));
 					}
-
+					this.getGermplasmDataManager().generateGermplasmUUID(cropType, Arrays.asList(germplasm));
```
darla-leafnode, line 343: Use `CollectionUtils.singletonList`

PR #890, opened 2020-09-21
IBP-4021 germplasm search api

nahuel-soldevilla:
Sharing the progress so far, the idea is to merge the current work for v16.1, so that at least the api is available in uat, and continue working on the pending items next sprint, probably in parallel with the frontend.

Implemented:
- Pedigree retrieval
- Almost all filters

TODO
- Filters
   - program
   - attributes
- Special columns (FGID, CROSS, etc)
- integration tests (tests for the existing search will be migrated to the new one at the end)
- other services: location info, etc
- possibly other minor details

Let me know if you have any concerns, I'll include a summary in Confluence later.

cc @clarysabel @darla-leafnode @abatac


src/main/java/org/generationcp/middleware/api/germplasm/search/GermplasmSearchResponse.java
```diff
@@ -0,0 +1,290 @@
...
+public class GermplasmSearchResponse {
+
+	// canonical columns
+
+	private int gid;
+	private Integer groupId;
+	private String names;
+	private String methodName;
+	private String locationName;
+	private String stockIds;
```
clarysabel, line 21: @nahuel-soldevilla Let's check the inventory fields with @mcrimi  
We are planning to remove stockIds in the current germplasm search and availableBalance was splitted in 2 columns in other places: availableBalance and unit. For the inventory, I think that the lotCount should be included here too. I will merge it anyway, it can be done in the next sprint

PR #897, opened 2020-10-05
Fix GET locations service and all places where it is used

clarysabel:
Logic: If program is null, then only locations with no program assigned are retrieved. If program is not null, then locations with no program assigned plus locations assigned to the current program are retrieved.
IBP-4049


cuenyad (approved):
Looks, good!

PR #901, opened 2020-10-13
IBP-2753-ImportKSUDateFormat

gelli-leafnode:
Added logic for checking if date value uses default pattern or ksu date pattern

### DIFF
--- a/src/main/java/org/generationcp/middleware/util/Util.java
+++ b/src/main/java/org/generationcp/middleware/util/Util.java
@@ -62,8 +62,6 @@ public class Util {
 	public static final String FRONTEND_DATE_FORMAT_3 = "dd/MM/yyyy";
 	public static final String FRONTEND_TIMESTAMP_FORMAT = "yyyy-MM-dd hh:mm:ss";
 	public static final String DATE_AS_NUMBER_FORMAT_KSU = "d/M/yy";
-	public static final Pattern ksuDatePattern = Pattern.compile("^(0?[1-9]|[12][0-9]|3[01])\\/(0?[1-9]|1[012])\\/((19|20)\\d\\d|\\d\\d)");
-	public static final Pattern defaultDatePattern = Pattern.compile("^((19|20)\\d\\d)(0?[1-9]|1[012])(0?[1-9]|[12][0-9]|3[01])");
 
 	private Util() {
 		// make a private constructor to hide the implicit public one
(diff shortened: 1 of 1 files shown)


src/main/java/org/generationcp/middleware/util/Util.java
```diff
@@ -490,4 +494,29 @@ public static boolean isLeapYear(final int year) {
...
 
+
+	public static boolean isDateMatchPattern(final String value, final String dateFormat) {
+		final String strPattern;
+		if (Util.DATE_AS_NUMBER_FORMAT_KSU.equals(dateFormat)) {// d/M/yy|yyyy
+			strPattern = "^(0?[1-9]|[12][0-9]|3[01])\\/(0?[1-9]|1[012])\\/((19|20)\\d\\d|\\d\\d)";
+		} else {// yyyyyMMdd
+			strPattern = "^((19|20)\\d\\d)(0?[1-9]|1[012])(0?[1-9]|[12][0-9]|3[01])";
+		}
+		final Pattern pattern = Pattern.compile(strPattern);
```
abatac, line 505: If this method is called in a loop, Pattern.compile would be costly. Move this to class variable so that pattern is only compiled once.

src/main/java/org/generationcp/middleware/util/Util.java
```diff
@@ -490,4 +497,27 @@ public static boolean isLeapYear(final int year) {
 		return isLeapYear;
 	}
 
+
+	public static boolean isDateMatchPattern(final String value, final String dateFormat) {
```
darla-leafnode, line 501: Rename as isDateMatchesPattern

src/main/java/org/generationcp/middleware/util/Util.java
```diff
@@ -490,4 +497,27 @@ public static boolean isLeapYear(final int year) {
...
+		final Matcher matcher;
+		if (Util.DATE_AS_NUMBER_FORMAT_KSU.equals(dateFormat)) {// d/M/yy|yyyy
+			matcher = ksuDatePattern.matcher(value);
+		} else {// yyyyyMMdd
+			matcher = defaultDatePattern.matcher(value);
+		}
+		return matcher.matches();
+	}
+
+	public static Date tryParseDateAccurately(final String date, final String format) {
```
darla-leafnode, line 511: "accurate" doesn't seem to apply here. Let's rename to tryParseDateByFormat

src/main/java/org/generationcp/middleware/util/Util.java
```diff
@@ -490,4 +497,27 @@ public static boolean isLeapYear(final int year) {
...
+		try {
+			if (Util.isDateMatchPattern(date, format)) {
```
darla-leafnode, line 513: I wonder if we need this check here if date matches pattern. Without it, there will be a parse/null exception which will return null anyway

gelli-leafnode: @darla-leafnode, 
I think this can be removed and just use Util.parseDate. This is originally meant to check pattern yyMMdd, some data passes this pattern even if it's pattern is dd/MM/yy and produced wrong output.

gelli-leafnode:
Closing this PR changes are not needed

darla-leafnode (approved):

darla-leafnode (approved):

darla-leafnode:
As discussed, please create a new PR with just the final change. I don't want to merge this branch's history

PR #903, opened 2020-10-14
IBP-4071 IBP-3939 Germplasm Search

abatac:


nahuel-soldevilla (changes requested):
My TODO:

- [x] Split Study of use filters (Comments in AC)
- [x] Include other progenitors
- [ ] integration test migration (Optional, can be done when deprecating new module)

PR #904, opened 2020-10-15
IBP-3528-ValidateTrialInstanceValue

gelli-leafnode:
Created reusable logic for checking if measurement variable datatype and it's value is correct (numeric) 
Added logic for checking trial instance value in observation variables

### DIFF
--- a/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
+++ b/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
@@ -1201,12 +1201,12 @@ public class DataImportServiceImpl extends Service implements DataImportService 
 		for (final MeasurementRow row : observations) {
 			for (final MeasurementData data : row.getDataList()) {
 				if (data.getMeasurementVariable().getTermId() == TermId.TRIAL_INSTANCE_FACTOR.getId()) {
-					final Message message = Util.validateVariableValues(data.getMeasurementVariable(), data.getValue());
-					if (message != null) {
+					final Optional<Message> message = Util.validateVariableValues(data.getMeasurementVariable(), data.getValue());
+					if (message.isPresent()) {
 						if (errors.containsKey(Constants.INVALID_TRIAL)) {
-							errors.get(Constants.INVALID_TRIAL).add(message);
+							errors.get(Constants.INVALID_TRIAL).add(message.get());
 						} else {
-							errors.put(Constants.INVALID_TRIAL, Arrays.asList(message));
+							errors.put(Constants.INVALID_TRIAL, Arrays.asList(message.get()));
 						}
 					}
 				}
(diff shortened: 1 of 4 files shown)


src/main/java/org/generationcp/middleware/util/Util.java
```diff
@@ -490,4 +495,21 @@ public static boolean isLeapYear(final int year) {
...
+	 */
+	public static Message validateVariableValues(final MeasurementVariable variable, final String value) {
+
+		if ((variable.getDataTypeId() != null && variable.getDataTypeId().equals(DataType.NUMERIC_VARIABLE.getId())) || variable.getDataType().equals(DataType.NUMERIC_VARIABLE.getDataTypeCode())) {
+			if (!StringUtils.isNumeric(value)) {
+				return new Message(Constants.INVALID_NUMERIC_VALUE_MESSAGE, variable.getLabel(), value);
+			}
+		}
+
+		return null;
```
abatac, line 512: return Optional.empty instead of null

src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -1189,4 +1197,30 @@ public int getMaxRowLimit() {
 	public void setMaxRowLimit(final int maxRowLimit) {
 		this.maxRowLimit = maxRowLimit;
 	}
+
+	private void validateTrialInstanceValueInObservation (final List<MeasurementRow> observations, final Map<String, List<Message>> errors) {
+		for (final MeasurementRow row : observations) {
+			for (final MeasurementData data : row.getDataList()) {
+				if (data.getMeasurementVariable().getTermId() == TermId.TRIAL_INSTANCE_FACTOR.getId()) {
+					final Optional<Message> message = Util.validateVariableValues(data.getMeasurementVariable(), data.getValue());
+					if (message.isPresent()) {
```
darla-leafnode, line 1206: if-else can be simplified as:
`errors.putIfAbsent(Constants.INVALID_TRIAL, new ArrayList<>());
					errors.get(Constants.INVALID_TRIAL).add(message.get());`

src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -1099,6 +1102,11 @@ public int saveProjectData(final Workbook workbook, final String programUUID, fi
 
 		this.checkForExistingTrialInstance(workbook, errors, programUUID);
 
+		// Validate Trial Instance Value
```
darla-leafnode, line 1105: As discussed, please check if you can put the numeric validation inside `checkForExistingTrialInstance` method (will have to rename to `validateTrialInstanceValues`)

darla-leafnode (approved):

PR #907, opened 2020-10-20
IBP-4029 convert germplasm tab to angular table

mderamos:


src/main/java/org/generationcp/middleware/domain/dms/Enumeration.java
```diff
@@ -65,6 +65,14 @@ public String getDescription() {
 		return this.description;
 	}
 
+	public int getRank(){
```
darla-leafnode, line 68: Is this change needed still?

mderamos: Hi @darla-leafnode, it's no longer needed. I'll just revert the commit.

darla-leafnode (approved):

PR #908, opened 2020-10-20
Ibp 4057 filter and sort entries table

lik-83:


src/main/java/org/generationcp/middleware/dao/dms/StockDao.java
```diff
@@ -439,4 +517,88 @@ private void addFixedVariableIfPresent(final TermId termId, final String value,
...
+				final String variableTypeString = filter.getVariableTypeMap().get(observableId);
+				this.applyFactorsFilter(sql, observableId, variableTypeString, performLikeOperation);
+			}
+		}
+	}
+
+	private void applyFactorsFilter(final StringBuilder sql, final String variableId, final String variableType,
+		final boolean performLikeOperation) {
+		// Check if the variable to be filtered is in one of the columns in stock, nd_experiment, geolocation or sum of samples
+		final String observationUnitClause = VariableType.OBSERVATION_UNIT.name().equals(variableType) ? "nde.observation_unit_no" : null;
```
clarysabel, line 541: seems to me that this block is not needed. we do not have observation unit, neither sample in this case @lik-83

lik-83: Done!

src/main/java/org/generationcp/middleware/dao/dms/StockDao.java
```diff
@@ -1,3 +1,4 @@
+
```
clarysabel, line 1: @lik-83 remove empty line

PR #909, opened 2020-10-21
IBP-4055-TrialHeaderAlias

gelli-leafnode:
When compairing trial headers, use measurement name instead of label to verify if header exists.


src/main/java/org/generationcp/middleware/domain/etl/MeasurementRow.java
```diff
@@ -95,7 +95,7 @@ public void setExperimentId(final int experimentId) {
 			for (final MeasurementData data : this.dataList) {
 				//Remove SAMPLES with TermId = -2 because is nonexistent term Id
 				final boolean isSampleVariable = data.getMeasurementVariable().getTermId() == TermId.SAMPLES.getId();
-				if ( (trialHeaders == null || !trialHeaders.contains(data.getLabel()) ) && !isSampleVariable) {
+				if ( (trialHeaders == null || !trialHeaders.contains(data.getMeasurementVariable().getName()) ) && !isSampleVariable) {
```
abatac, line 98: Hi @gelli-leafnode why do need to change this to variable name?

gelli-leafnode: Hi @abatac , 
Currently the trial header list contains the trialvariable names, when the user maintain the header mapping the label value doesn't match the contents of trial header.

PR #912, opened 2020-10-26
IBP-3944 save germplasm list

nahuel-soldevilla:
- get germplasm list types
- save germplasm list 
- new enum for GermplasmList Status

No new permission created for now


src/main/java/org/generationcp/middleware/api/germplasmlist/GermplasmListGeneratorDTO.java
```diff
@@ -0,0 +1,144 @@
...
+
+public class GermplasmListGeneratorDTO {
+
+	public static class GermplasmEntryDTO {
+
+		private Integer entryNo;
+		private Integer gid;
+		private String entryCode;
+		private String seedSource;
+		private String designation;
```
clarysabel, line 16: @nahuel-soldevilla I think we should not allow sending the designation through the entries, I guess it should always be calculated based on the preferred name

clarysabel (commented):
Leaving some comments

PR #913, opened 2020-10-28
Ibp 4121 import germplasm services

clarysabel:
Created on @lik-83 behalf


src/main/java/org/generationcp/middleware/api/breedingmethod/BreedingMethodServiceImpl.java
```diff
@@ -44,4 +46,27 @@ public BreedingMethodDTO getBreedingMethod(final Integer breedingMethodDbId) {
...
 	}
+
+	@Override
+	public List<BreedingMethodDTO> getBreedingMethods(final String programUUID, final boolean favorites) {
+		final List<Method> breedingMethods;
+		if (!StringUtils.isEmpty(programUUID) && favorites) {
+			final List<Integer> breedingMethodIds = this.getFavoriteProjectMethodsIds(programUUID);
+			breedingMethods = this.daoFactory.getMethodDAO().getMethodsByIds(breedingMethodIds);
+		} else {
+			breedingMethods = this.daoFactory.getMethodDAO().getAllMethod();
```
clarysabel, line 57: Almost sure that this will not cover the case where an user created a method for a program and then he wants to know all the breeding methods, I think this query will exclude the ones that were created for an specific program

src/test/java/org/generationcp/middleware/manager/GermplasmDataManagerIntegrationTest.java
```diff
@@ -1307,6 +1309,24 @@ public void testGenerateGermplasmUUIDUseUUIDFalse() {
 		Assert.assertEquals(20, germplasm.getGermplasmUUID().length());
 	}
 
+	@Test
+	public void shouldGetUserDefinedFieldByTableTypeAndCodes() {
+
```
clarysabel, line 1314: please use enum

clarysabel (commented):
Leaving some comments @lik-83

PR #914, opened 2020-10-29
IBP-4120 germplasm template

cuenyad:
Please review.

Include:
https://github.com/IntegratedBreedingPlatform/BMSAPI/pull/419


src/main/java/org/generationcp/middleware/dao/UserDefinedFieldDAO.java
```diff
@@ -37,12 +38,12 @@
 	private static final Logger LOG = LoggerFactory.getLogger(UserDefinedFieldDAO.class);
 
 	@SuppressWarnings("unchecked")
-	public List<UserDefinedField> getByFieldTableNameAndType(final String tableName, final String fieldType) {
+	public List<UserDefinedField> getByFieldTableNameAndType(final String tableName, final Set<String> fieldType) {
```
clarysabel, line 41: Even when we may end not using this function I am in favor of using Set instead of String :+1:

PR #915, opened 2020-11-04
IBP-4085 add germplasm to study list

mderamos:


src/main/java/org/generationcp/middleware/dao/dms/StockDao.java
```diff
@@ -117,6 +117,21 @@ public long countStudiesByGid(final int gid)  {
 		}
 	}
 
+	public boolean hasUnassignedStudyEntriesToPlot(final int studyId) {
+		try {
+			final SQLQuery query = this.getSession()
+				.createSQLQuery("SELECT COUNT(*) FROM ibdbv2_wheat_merged.stock s "
```
clarysabel, line 123: @mderamos I think ibdbv2_wheat_merged is not needed

src/main/java/org/generationcp/middleware/dao/dms/StockDao.java
```diff
@@ -335,7 +350,24 @@ public void replaceExperimentStocks(final Integer oldStockId, final Integer newS
 			LOG.error(errorMessage, e);
 			throw new MiddlewareQueryException(errorMessage, e);
 		}
+	}
 
+	public Integer getNextEntryNumber(final Integer studyId) {
+		try {
+			final String queryString = "SELECT MAX(Convert(s.uniquename, SIGNED)) FROM stock s where s.project_id = :studyId";
```
clarysabel, line 357: @mderamos would not be clear to use IFNULL in the query instead og having a IF in java? also you can sum + 1 in the query
something like 
`SELECT IFNULL(MAX(Convert(s.uniquename, SIGNED)), 0) + 1 FROM stock s where s.project_id = :studyId`

src/main/java/org/generationcp/middleware/service/impl/study/StudyEntryServiceImpl.java
```diff
@@ -170,6 +175,11 @@ public void updateStudyEntryProperty(final int studyId, final StudyEntryProperty
 		}
 	}
 
+	@Override
+	public Boolean hasUnassignedStudyEntriesToPlot(final int studyId) {
```
clarysabel, line 179: @mderamos I found this name hard to read, is it the same than asking if the study has observations generated?

src/main/java/org/generationcp/middleware/domain/study/StudyEntryGeneratorBatchRequestDto.java
```diff
@@ -0,0 +1,41 @@
+package org.generationcp.middleware.domain.study;
+
+import org.generationcp.middleware.domain.inventory.common.SearchCompositeDto;
+import org.pojomatic.annotations.AutoProperty;
+
+@AutoProperty
+public class StudyEntryGeneratorBatchRequestDto {
```
clarysabel, line 7: I would remove Batch from the name, given that we are using Generator @mderamos

src/main/java/org/generationcp/middleware/service/impl/study/StudyEntryServiceImpl.java
mderamos: Hi @clarysabel, it's used for showing a warning when the study has a generated design and the user adds a new study entries. The added entries are not assigned to plots yet.

src/main/java/org/generationcp/middleware/domain/study/StudyEntryListGeneratorRequestDto.java
```diff
@@ -0,0 +1,18 @@
+package org.generationcp.middleware.domain.study;
+
+import org.pojomatic.annotations.AutoProperty;
+
+@AutoProperty
+public class StudyEntryListGeneratorRequestDto {
```
clarysabel, line 6: @mderamos does it make sense to create a class only to have one attribute? Could not be possible for the service to receive an Integer instead of this class?

src/main/java/org/generationcp/middleware/service/impl/study/StudyEntryServiceImpl.java
clarysabel: @mderamos understood, in that case I think we should rename it to hasNotAssignedEntries or something simpler. If you still want to use Plot I suggest to use ObservationUnit instead but hasNotAssignedEntries is shorter and at the end there is no other object to assign entries besides the Observation Unit so it should be ok just make it shorter

darla-leafnode (approved):

PR #917, opened 2020-11-04
Ibp 4097 import germplasm backend

clarysabel:


src/main/java/org/generationcp/middleware/api/germplasm/GermplasmGuidGenerator.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+import java.util.UUID;
+
+public class GermplasmGuidGenerator {
+
+	public static final String MID_STRING = "G";
+	public static final int SUFFIX_LENGTH = 15;
+
+	public static void generateGermplasmGuids(final CropType crop, final List<Germplasm> germplasmList) {
+		Preconditions.checkNotNull(crop);
```
nahuel-soldevilla, line 18: This looks interesting. I wonder if it could be abstracted even more. Something like this:
```java
// maybe in org.generationcp.middleware.util.Util?
public static <T extends UIDAssignable> void generateUIDs(
	final CropType crop, final List<T> list,
	final String midString,
	final Object suffixLength) {

	final boolean doUseUUID = crop.isUseUUID();
	for (final T item : list) {
		if (item.getUID() == null) {
			if (doUseUUID) {
				item.setUID(UUID.randomUUID().toString());
			} else {
				final String cropPrefix = crop.getPlotCodePrefix();
				item.setUID(cropPrefix + midString + randomAlphanumeric(suffixLength));
			}
		}
	}
}
...
private static interface UIDAssignable {
	void setUID(String uid);
	String getUID();
}

```

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmServiceImpl.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Service
+@Transactional
+public class GermplasmServiceImpl implements GermplasmService {
+
+	private final DaoFactory daoFactory;
+
+	private static final String DEFAULT_DASH = "-";
```
nahuel-soldevilla, line 39: Maybe rename to DEFAULT_BIBREF_FIELD

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmServiceImpl.java
```diff
@@ -0,0 +1,166 @@
...
+		final Set<String> locationAbbreviations = germplasmDtos.stream().map(g -> g.getLocationAbbr()).collect(Collectors.toSet());
+		return this.daoFactory.getLocationDAO().getByAbbreviations(new ArrayList<>(locationAbbreviations)).stream()
+			.collect(Collectors.toMap(Location::getLabbr, Location::getLocid));
+	}
+
+	private Map<String, Method> getBreedingMethodsMapByAbbr(final List<GermplasmImportRequestDto.GermplasmDto> germplasmDtos) {
+		final Set<String> breedingMethods = germplasmDtos.stream().map(g -> g.getBreedingMethodAbbr()).collect(Collectors.toSet());
+		return this.daoFactory.getMethodDAO().getByCode(new ArrayList<>(breedingMethods)).stream()
+			.collect(Collectors.toMap(Method::getMcode, method -> method));
+	}
```
nahuel-soldevilla, line 138: Some of these methods look useful, maybe we can have them in the daos

nahuel-soldevilla (approved):
Looks good

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmServiceImpl.java
clarysabel: done

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmServiceImpl.java
clarysabel: Agree that may be useful but they depend on the user input, so I would not located in the DAO's. I saw in Aldrin's branch similar code, so when bot are merged we can find a way to reuse them

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmGuidGenerator.java
clarysabel: I have create https://ibplatform.atlassian.net/browse/IBP-4167

PR #919, opened 2020-11-05
New germplasm search known issues IBP-4114 IBP-4125 IBP-4115 IBP-4162

nahuel-soldevilla:


src/main/java/org/generationcp/middleware/service/pedigree/PedigreeServiceImpl.java
```diff
@@ -147,7 +150,20 @@ public String getCropName() {
...
+
+		final Iterable<List<Integer>> partition = Iterables.partition(gids, MAX_GID_LIST_SIZE);
+		final Map<Integer, String> crossExpansions = new HashMap<>();
+
+		for (final List<Integer> partitionedGidList : partition) {
+			crossExpansions.putAll(this.getCrossExpansions(new HashSet<>(partitionedGidList), level, crossExpansionProperties));
+		}
+		return crossExpansions;
+	}
+
```
nahuel-soldevilla, line 166: @IntegratedBreedingPlatform/developers 
I've copied this common workaround directly to the pedigree service, so that we don't have to replicate every time we need it. Let me know if you have any concern with this.

PR #923, opened 2020-11-17
IBP-4174 reduce  Germplsm UUID length

mderamos:


darla-leafnode:
Can you create another branch/PR with just a single commit for the final changeset/s for this? I don't want to merge the commit history with all the reverts.

PR #925, opened 2020-11-18
IBP-4174 change germplsm UUID length

mderamos:


nahuel-soldevilla:
Seems fine to me. Is there any reason why you chose 15 originally? The previous scripts (obs_unit_id, plot_uuid) use 8 too.

Given that these scripts take some time run (1 hour according to the last test), wouldn't make sense to only keep this version and delete the previous changesets? (both uuid migration changesets from 16_1_0.xml, and deleted germplasm migration from 16_3_0.xml)
cc @clarysabel 

---
Also, if you are using Intellij, there's a setting that allows you to paste text with whitespaces into xml (e.g sql):
intellij-keep-whitespaces

src/main/resources/liquibase/crop_changelog/16_4_0.xml
```diff
@@ -0,0 +1,225 @@
...
+				SELECT
+				COUNT(*)
+				FROM
+				germplsm
+				WHERE
+				germplsm_uuid IS NULL;
+			</sqlCheck>
+		</preConditions>
+		<comment>
+			Set germplsm_uuid values to NULL
```
darla-leafnode, line 19: Can you expand on this comment here that: 
> This is for the case when a previous 16.1 changeset was deployed to test/dev environments. This should not run for production deployments

src/main/resources/liquibase/crop_changelog/16_4_0.xml
```diff
@@ -0,0 +1,225 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+				   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+	<changeSet author="mderamos" id="v16.4.0-1">
+		<preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="0">
+				SELECT
```
darla-leafnode, line 10: I think it's more intuitive and easier to understand to check if there are any records that have **not null** germplasm_uuid values.
```
<sqlCheck expectedResult="1">
				SELECT COUNT(*) > 0
				FROM germplsm
				WHERE germplsm_uuid IS NOT NULL;
			</sqlCheck>
```

src/main/resources/liquibase/crop_changelog/16_4_0.xml
```diff
@@ -0,0 +1,225 @@
...
+				germplsm
+				WHERE
+				germplsm_uuid IS NULL;
+			</sqlCheck>
+		</preConditions>
+		<comment>
+			Set germplsm_uuid values to NULL
+		</comment>
+		<sql dbms="mysql" splitStatements="false">
+			UPDATE germplsm SET germplsm_uuid = NULL;
```
darla-leafnode, line 22: To be sure, lets add a WHERE clause on the update statement:
```
UPDATE germplsm 
SET germplsm_uuid = NULL
where germplsm_uuid IS NOT NULL;
```

darla-leafnode (approved):

PR #927, opened 2020-11-20
IBP-3891: get preferred name from names if it's not present

lik-83:


lik-83:
This change is no longer needed.

PR #929, opened 2020-11-24
IBP-3891: added method to get LotStatus by name

lik-83:


lik-83:
Closed because it's no longer needed.

PR #932, opened 2020-11-27
IBP-4150 Germplasm Search improvements

darla-leafnode:


src/main/java/org/generationcp/middleware/dao/UserDefinedFieldDAO.java
```diff
@@ -155,4 +155,31 @@ public UserDefinedField getByTableTypeAndCode(final String table, final String t
 		return returnList;
 	}
 
+	public List<org.generationcp.middleware.api.nametype.GermplasmNameTypeDTO> searchNameTypes(final String query) {
+		if (StringUtils.isBlank(query)) {
+			return Collections.EMPTY_LIST;
```
abatac, line 160: Replace Collections.EMPTY_LIST with Collections.emptyList()

PR #935, opened 2020-11-30
Ibp 4098 import germplasm

clarysabel:


src/main/java/org/generationcp/middleware/dao/GermplasmDAO.java
```diff
@@ -110,6 +115,36 @@
...
+		+ "    g.germplsm_uuid as germplasmUUID, " //
+		+ "    n.nval as preferredName, " //
+		+ "    cast(g.gdate as char) as creationDate, " //
+		+ "    r.analyt as reference, " //
+		+ "    l.locid as breedingLocationId, " //
+		+ "    l.lname as breedingLocation, " //
+		+ "    m.mid as breedingMethodId, " //
+		+ "    m.mname as breedingMethod, " //
+		+ "    if(g.mgid > 0, true, false) as isGroupedLine, " //
+		+ "    g.mgid as groupId ";
```
nahuel-soldevilla, line 129: `isGroupedLine` logic could also be in the dto getter I guess

src/main/resources/liquibase/crop_changelog/16_4_0.xml
```diff
@@ -221,4 +221,17 @@
...
+				<indexExists tableName="germplsm" indexName="germplsm_idx14"></indexExists>
+			</not>
+		</preConditions>
+		<comment>Add index for germplsm gid and germplsm_uuid columns</comment>
+		<createIndex indexName="germplsm_idx14" tableName="germplsm" unique="false">
+			<column name="gid"></column>
+			<column name="germplsm_uuid"></column>
+		</createIndex>
+	</changeSet>
+
```
nahuel-soldevilla, line 236: Looks fine. Did you have time to investigate in which way it improves the performance? The only clause I can think of is the [`OR`](https://github.com/IntegratedBreedingPlatform/Middleware/pull/935/files#diff-70aa9494d2c25b3565e0b4a4580ccb910bd79b83399e6f82bf718d5c6ea1176aR1786) case, but it's strange because I would expect it to use the normal indices in that case. But if it works..

nahuel-soldevilla (commented):
Matches looks fine to me @clarysabel

nahuel-soldevilla:
continues on sprint 17.1

PR #940, opened 2020-12-03
IBP-4164 Get the term name of entry number to be used for sorting

mderamos:


src/main/java/org/generationcp/middleware/service/impl/study/StudyEntryServiceImpl.java
```diff
@@ -50,9 +55,11 @@ public StudyEntryServiceImpl(final HibernateSessionProvider sessionProvider) {
 
 	@Override
 	public List<StudyEntryDto> getStudyEntries(final int studyId) {
+		// Get entry number term. Name will be used for sorting
+		final Term entryNumberTerm = this.ontologyDataManager.getTermById(Integer.valueOf(TermId.ENTRY_NO.getId()));
 		// Return by ascending order of entry number. We need to perform cast first on uniquename since it's stored as string
```
darla-leafnode, line 60: We can remove this comment, right?
`// Return by ascending order of entry number. We need to perform cast first on uniquename since it's stored as string`

darla-leafnode (approved):

PR #946, opened 2020-12-11
IBP-4224: don't return studies if they were deleted

lik-83:


src/main/java/org/generationcp/middleware/dao/dms/DmsProjectDao.java
```diff
@@ -203,6 +203,7 @@
 		+ " WHERE "
 		+ "     nde.type_id = " + TermId.TRIAL_ENVIRONMENT_EXPERIMENT.getId()
 		+ "         AND geoloc.nd_geolocation_id = :instanceId "
+		+ "         AND proj.deleted = 0 "
```
clarysabel, line 206: @lik-83 could you use a constant? a new one or DELETED_STUDY ?

PR #949, opened 2020-12-15
IBP-4069-AddEntryTypeIfNotExists

gelli-leafnode:
Added addEntryTypeVariableIfNotExists. Added default value 'Test Entry' to measurement row


src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -429,6 +430,28 @@ public void processExperimentalDesign(final Workbook workbook, final String prog
 
 	}
 
+	@Override
+	public void addEntryTypeVariableIfNotExists(final Workbook workbook, final List<MeasurementVariable> measurementVariables,
+		final String programUUID) {
+		final List<MeasurementVariable> combinedVariables = new ArrayList<>();
+		combinedVariables.addAll(workbook.getConditions());
```
mderamos, line 437: You're adding the created ENTRY_TYPE variable in the factors list so I'm assuming that if it already exists, it's stored in the factors list. Are there instances where the ENTRY_TYPE variable is found in the conditions list?

src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -429,6 +430,25 @@ public void processExperimentalDesign(final Workbook workbook, final String prog
...
+		final boolean entryTypeIdExists = this.findMeasurementVariableByTermId(TermId.ENTRY_TYPE.getId(), workbook.getFactors()).isPresent();
+
+		// If ENTRY_TYPE variable is not existing in both Condition and Factors Section of workbook
+		// Automatically add ENTRY_TYPE variable as it is required in creating a new Study.
+		if (!entryTypeIdExists) {
+			final MeasurementVariable entryType = this.createMeasurementVariable(TermId.ENTRY_TYPE.getId(), null, Operation.ADD, PhenotypicType.GERMPLASM, programUUID);
+			workbook.getFactors().add(entryType);
+			// Add ENTRY_TYPE variable in Measurement Row with default value Test Entry
+			for (final MeasurementRow row : workbook.getObservations()) {
+				final MeasurementData data = new MeasurementData(entryType.getLabel(), String.valueOf(SystemDefinedEntryType.TEST_ENTRY.getEntryTypeCategoricalId()), false, entryType.getDataType(), entryType);
```
darla-leafnode, line 446: Is it possible to create MeasurementData outside of loop? If so we can use forEach clause to simplify the code:
`final MeasurementData data = new MeasurementData(entryType.getLabel(), String.valueOf(SystemDefinedEntryType.TEST_ENTRY.getEntryTypeCategoricalId()), false, entryType.getDataType(), entryType);
			workbook.getObservations().forEach(row -> row.getDataList().add(data));`

darla-leafnode (approved):

PR #955, opened 2020-12-18
IBP-4255-AddEntryTypeSucceedingImport

gelli-leafnode:
Modify logic in addEntryTypeVariableIfNotExists, if Entry_Type exists do not create the variable, if not exists in observation add default value in observation.
cc @darla-leafnode

### DIFF
--- a/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
+++ b/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
@@ -1253,7 +1253,7 @@ public class DataImportServiceImpl extends Service implements DataImportService 
 				rowWithoutTermid.add(row);
 			}
 		}
-		if (Util.isEmpty(rowList)) {
+		if (Util.isEmpty(rowWithoutTermid)) {
 			return Optional.absent();
 		} else {
 			return Optional.of(rowWithoutTermid);
--- a/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
+++ b/src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
@@ -61,6 +61,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 @Transactional
 public class DataImportServiceImpl extends Service implements DataImportService {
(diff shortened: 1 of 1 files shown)


src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -434,13 +435,19 @@ public void processExperimentalDesign(final Workbook workbook, final String prog
 	public void addEntryTypeVariableIfNotExists(final Workbook workbook,
 		final String programUUID) {
 
-		final boolean entryTypeIdExists = this.findMeasurementVariableByTermId(TermId.ENTRY_TYPE.getId(), workbook.getFactors()).isPresent();
+		final Optional<MeasurementVariable> optEntryType = this.findMeasurementVariableByTermId(TermId.ENTRY_TYPE.getId(), workbook.getFactors());
```
darla-leafnode, line 438: Rename as `entryTypeFactor`

src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -434,13 +435,19 @@ public void processExperimentalDesign(final Workbook workbook, final String prog
...
+		if (!optEntryType.isPresent() || !entryTypeidExistsInRow) {
+			final MeasurementVariable entryType;
+			if (!optEntryType.isPresent()) {
+				// If ENTRY_TYPE variable is not existing in Factors Section of workbook
+				// Automatically add ENTRY_TYPE variable as it is required in creating a new Study.
+				entryType = this.createMeasurementVariable(TermId.ENTRY_TYPE.getId(), null, Operation.ADD, PhenotypicType.GERMPLASM, programUUID);
+				workbook.getFactors().add(entryType);
+			} else {
+				entryType = optEntryType.get();
+			}
```
darla-leafnode, line 450: What if there are only some, not all, rows that don't have ENTRY_TYPE value? All of them will be overwritten to have TEST entry type, right? (if so, that is not correct behavior)

darla-leafnode: Better to find only the rows without ENTRY_TYPE data and update those to have TEST_ENTRY value

src/main/java/org/generationcp/middleware/service/DataImportServiceImpl.java
```diff
@@ -1246,4 +1239,24 @@ private void setMeasurementIdIfNecessary(final List<MeasurementVariable> measure
...
+	 */
+	private Optional<List<MeasurementRow>> findMeasurementRowWithoutTermId(final Integer termId, final List<MeasurementRow> rowList) {
+		final List<MeasurementRow> rowWithoutTermid = new ArrayList<>();
+		for (final MeasurementRow row: rowList) {
+			final boolean hasTermId = row.getDataList().stream().filter(measurementData -> measurementData.getMeasurementVariable().getTermId() == termId).collect(Collectors.toList()).size() > 0;
+			if (!hasTermId) {
+				rowWithoutTermid.add(row);
+			}
+		}
+		if (Util.isEmpty(rowList)) {
```
darla-leafnode, line 1256: this should be Util.isEmpty(**rowWIthoutTermId**)?

darla-leafnode (approved):

PR #958, opened 2020-12-28
IBP-2778: improve UserResource error handling

lik-83:


src/main/java/org/generationcp/middleware/service/impl/user/UserServiceImpl.java
```diff
@@ -135,54 +140,52 @@ public Integer createUser(final UserDto userDto) {
...
+
+			final Set<CropType> crops = new HashSet<>();
+			for (final CropDto crop : userDto.getCrops()) {
+				final CropType cropType = new CropType(crop.getCropName());
+				crops.add(cropType);
+			}
+			person.setCrops(crops);
 
 			final WorkbenchUser recordSaved = this.workbenchDaoFactory.getWorkbenchUserDAO().saveOrUpdate(user);
 			idUserSaved = recordSaved.getUserid();
```
clarysabel, line 175: @lik-83 we dont need this variable right? we can return recordSaved.getUserid() at the end of the try block

src/main/java/org/generationcp/middleware/service/impl/user/UserServiceImpl.java
```diff
@@ -129,60 +134,57 @@ public long countAllUsers() {
 	@Override
 	public Integer createUser(final UserDto userDto) {
 
-		Integer idUserSaved = null;
+//		Integer idUserSaved = null;
```
clarysabel, line 137: @lik-83 remove comment

lik-83: Done!

PR #962, opened 2021-01-05
IBP-4138 Removing rank column from stockprop table.

cuenyad:
Please, review.

### DIFF
--- a/src/main/resources/liquibase/crop_changelog/17_1_0.xml
+++ b/src/main/resources/liquibase/crop_changelog/17_1_0.xml
@@ -1,0 +1,23 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+				   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+	<changeSet author="diego" id="v17_1_0">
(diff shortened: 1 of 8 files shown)


nahuel-soldevilla (approved):

src/main/resources/liquibase/crop_changelog/17_1_0.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+				   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+	<changeSet author="diego" id="v17.1.0-1">
+		<preConditions onFail="MARK_RAN">
+			<and>
```
clarysabel, line 9: and seems to be not required here @cuenyad

src/main/resources/liquibase/crop_changelog/17_1_0.xml
```diff
@@ -0,0 +1,23 @@
...
+				<columnExists tableName="stockprop" columnName="rank"></columnExists>
+			</and>
+		</preConditions>
+		<comment>Remove rank column from stockprop table</comment>
+		<dropIndex tableName="stockprop" indexName="stockprop_idx1"></dropIndex>
```
clarysabel, line 14: According to MySQL documentation, dropping the index should not be required to drop a column. Have you tested that before? See: https://stackoverflow.com/questions/4341897/what-happens-if-i-drop-a-mysql-column-without-dropping-its-index-first 
@cuenyad

PR #963, opened 2021-01-05
 IBP-3935 Removing previousAmount column from ims_transaction table.

cuenyad:
Please review.

Included:
https://github.com/IntegratedBreedingPlatform/Workbench/pull/412


src/main/resources/liquibase/crop_changelog/17_1_0.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+				   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+
+	<changeSet author="diego" id="v17.1.0-2">
+		<preConditions onFail="MARK_RAN">
+			<and>
```
clarysabel, line 10: I think this and is not required right? because there is only one precondition @cuenyad

PR #967, opened 2021-01-08
Ibp 2629 remove database name in database broker

clarysabel:
Delete DatabaseBroker
Cleanup StudyService


src/main/java/org/generationcp/middleware/operation/destroyer/StudyDestroyer.java
```diff
@@ -1,42 +1,46 @@
...
 
+import java.util.List;
+
 public class StudyDestroyer extends Destroyer {
 
+	private DaoFactory daoFactory;
+
 	public StudyDestroyer(final HibernateSessionProvider sessionProviderForLocal) {
 		super(sessionProviderForLocal);
+		this.daoFactory = new DaoFactory(sessionProvider);
```
nahuel-soldevilla, line 17: We can also remove Destroyer class I think

src/main/java/org/generationcp/middleware/manager/UserProgramStateDataManagerImpl.java
```diff
@@ -1,28 +1,29 @@
...
  */
 @Transactional
 public class UserProgramStateDataManagerImpl extends DataManager implements UserProgramStateDataManager {
 
+	private DaoFactory daoFactory;
+
 	public UserProgramStateDataManagerImpl(HibernateSessionProvider sessionProvider) {
 		super(sessionProvider);
+		this.daoFactory = new DaoFactory(sessionProvider);
 	}
```
nahuel-soldevilla, line 27: It seems we can also cut ties with DataManager here, the only link between the two is the logAndThrowException method. Something similar may apply for other classes extending from DataManager

nahuel-soldevilla (approved):
Looks great!
I saw some things regarding DataManager and Destroyer, but we can include those in future deprecation efforts involving the data manager classes.

clarysabel:
@nahuel-soldevilla we still need to remove some other classes and migrate some SQL code from DataManagers to DAO so we dont use the session in DataManagers anymore. I will create a new ticket for that and I think it can be included sprint by sprint instead of waiting for the next technical period

PR #976, opened 2021-02-01
IBP 4109 Update Germplasm Progenitors and Breeding Method

abatac:


src/main/java/org/generationcp/middleware/api/germplasm/GermplasmServiceImpl.java
```diff
@@ -320,26 +333,113 @@ private void updateGermplasm(final Germplasm germplasm, final GermplasmUpdateDTO
 				Optional.ofNullable(Integer.parseInt(germplasmUpdateDTO.getCreationDate()));
```
darla-leafnode, line 333: Can just be `Optional.of(Integer.parseInt(germplasmUpdateDTO.getCreationDate()))`;

src/main/java/org/generationcp/middleware/dao/GermplasmDAO.java
```diff
@@ -497,6 +497,31 @@ public Germplasm getByGIDWithPrefAbbrev(final Integer gid) {
 		return new ArrayList<>();
 	}
 
+	/**
+	 * Only returns germplasm with progeny (Germplasm Id is assigned to other germplasm's gpid1, gpid2 OR progntrs.pid)
+	 *
+	 * @param gids - gids to be checked for descendants
+	 * @return
+	 */
+	public List<Germplasm> getGermplasmWithDescendants(final Set<Integer> gids) {
```
darla-leafnode, line 506: As commented in service seems we can only return GIDs instead of the whole entity

darla-leafnode (changes requested):

darla-leafnode (approved):

PR #989, opened 2021-02-19
IBP-4338-BrapiGetProgramsV2

gelli-leafnode:
Created ProgramSearchRequest to handle program filter replacement to pojo ProgramFilters (Map)


src/main/java/org/generationcp/middleware/service/api/program/ProgramSearchRequest.java
```diff
@@ -0,0 +1,42 @@
...
+        this.programDbId = programDbId;
+    }
+
+    public String getProgramName() {
+        return this.programName;
+    }
+
+    public void setProgramName(final String programName) {
+        this.programName = programName;
+    }
```
darla-leafnode, line 40: Implement hashCode, equals, toString

src/main/java/org/generationcp/middleware/manager/api/WorkbenchDataManager.java
```diff
@@ -65,7 +64,7 @@
 	 * @param filters   - the filters that to be included in the query
 	 * @return All projects based on the given start, numOfRows and filters Map
 	 */
-	List<Project> getProjects(final int start, final int numOfRows, final Map<ProgramFilters, Object> filters);
```
darla-leafnode, line 68: Use Pageable interface

PR #993, opened 2021-02-25
IBP-4207 Delete Germplasm

abatac:


src/main/java/org/generationcp/middleware/domain/germplasm/GermplasmDeleteResponse.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import org.pojomatic.Pojomatic;
+import org.pojomatic.annotations.AutoProperty;
+
+import java.util.Set;
+
+@AutoProperty
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class GermplasmDeleteResponse {
```
darla-leafnode, line 11: I think this class makes sense to be placed in BMSAPI since it is used there and not in MW?

darla-leafnode (approved):

PR #995, opened 2021-03-02
Ibp 4324 angular nav bar 3

lik-83:


src/main/java/org/generationcp/middleware/service/api/workbench/WorkbenchService.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public interface WorkbenchService {
+
+	WorkbenchSidebarCategoryLink getWorkbenchSidebarLinksByCategoryId(Integer workbenchSidebarCategoryLink);
+
+	void close();
```
clarysabel, line 13: is this close() required? @lik-83

src/main/resources/liquibase/workbench_changelog/17_2_0.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+				   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+	<changeSet author="lkovacic" id="v17.1.0-1">
```
clarysabel, line 7: @lik-83 id should be v17.2.0-2 and version 17.2

src/main/java/org/generationcp/middleware/service/api/user/UserDto.java
```diff
@@ -31,6 +31,8 @@
 	private String password;
 	private Set<CropDto> crops;
 	private Set<String> authorities;
+	private String selectedCropName;
```
clarysabel, line 34: @lik-83 why is this required? is it really an attribute of UserDto?

lik-83: These attributes are used to show information about the last opened project. Maybe, we can think to create an API to update and get that info.

src/main/java/org/generationcp/middleware/service/api/workbench/WorkbenchService.java
lik-83: done!

src/main/resources/liquibase/workbench_changelog/17_2_0.xml
lik-83: Done!

src/main/java/org/generationcp/middleware/service/api/user/UserDto.java
clarysabel: created: https://ibplatform.atlassian.net/browse/IBP-4397

PR #998, opened 2021-03-15
IBP-4447 Fix locations filter by program UUID

darla-leafnode:


clarysabel:
HI @darla-leafnode seems to be fine. I switched to your branch and run the tests (where they were failing) and now they are passing.

abatac (approved):

PR #999, opened 2021-03-16
IBP-4415 Germplasm Attribute Services

mderamos:


src/main/java/org/generationcp/middleware/domain/germplasm/GermplasmDto.java
```diff
@@ -34,106 +34,106 @@
 
 	private List<GermplasmNameDto> names;
 
-	private List<GermplasmAttributeDto> attributes;
+	private List<GermplasmAttributeRequestDto> attributes;
```
clarysabel, line 37: why is this class being replaced? to me makes more sense to use GermplasmAttributeDto when building a new GermplasmDto

clarysabel (changes requested):
HI @mderamos I am leaving some comments to your PR. 
I'll try to check BMSAPI later today

PR #1009, opened 2021-03-24
IBP-4052 new germplasm details popup

abatac:


src/main/java/org/generationcp/middleware/api/germplasm/GermplasmAttributeServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+		newAttribute.setAdate(Integer.valueOf(dto.getDate()));
+		return this.daoFactory.getAttributeDAO().save(newAttribute).getAid();
+	}
+
+	@Override
+	public void updateGermplasmAttribute(final Integer attributeId, final GermplasmAttributeRequestDto dto) {
+		final Attribute attribute = this.daoFactory.getAttributeDAO().getById(attributeId);
+		attribute.setAval(dto.getValue());
+		attribute.setLocationId(dto.getLocationId());
+		attribute.setAdate(Integer.valueOf(dto.getDate()));
```
darla-leafnode, line 47: @mcrimi - do we update the `atributs.adate` when updating the attribute?

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmServiceImpl.java
```diff
@@ -1161,6 +1165,63 @@ public long countGermplasmByStudy(final Integer studyDbId) {
...
+					.collect(Collectors.toMap(GermplasmDto::getGid, g -> g));
+			final GermplasmDto femaleParent = germplasmDtoMap.get(germplasmDto.getGpid1());
+			final List<GermplasmDto> maleParents = new ArrayList<>();
+			maleParentsGids.forEach(m -> {
+				if (germplasmDtoMap.containsKey(m)) {
+					maleParents.add(germplasmDtoMap.get(m));
+				}
+			});
+
+			if ("GEN".equals(method.getMtype())) {
```
darla-leafnode, line 1213: Use `MethodType.GENERATIVE.getCode()`

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmService.java
```diff
@@ -112,4 +114,9 @@
 
 	long countAttributesByGUID(String gemrplasmUUID, List<String> attributeDbIds);
 
+	List<GermplasmNameTypeDTO> filterGermplasmNameTypes(Set<String> codes);
```
darla-leafnode, line 117: It doesn't make intuitive sense to me to offer this as part of `GermplasmService` as it's not action to a specific (or group of) actual germplasm.

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmServiceImpl.java
```diff
@@ -1161,6 +1165,63 @@ public long countGermplasmByStudy(final Integer studyDbId) {
...
+
+	@Override
+	public ProgenitorsDetailsDto getGermplasmProgenitorDetails(final Integer gid) {
+		final GermplasmDto germplasmDto = this.daoFactory.getGermplasmDao().getGermplasmDtoByGid(gid);
+		if (germplasmDto != null) {
+			final List<Progenitor> progenitors = this.daoFactory.getProgenitorDao().getByGID(gid);
+			germplasmDto.setOtherProgenitors(progenitors.stream().map(p -> p.getProgenitorGid()).collect(Collectors.toList()));
+			final Method method = this.daoFactory.getMethodDAO().getById(germplasmDto.getBreedingMethodId());
+
+			final ProgenitorsDetailsDto progenitorsDetailsDto = new ProgenitorsDetailsDto();
```
darla-leafnode, line 1188: Create constructor with fields to set

src/main/java/org/generationcp/middleware/dao/GermplasmDAO.java
```diff
@@ -1710,4 +1716,38 @@ public boolean hasExistingCrosses(final Integer femaleParent, final List<Integer
 		}
 	}
 
+	public GermplasmDto getGermplasmDtoByGid(final Integer gid) {
+		final StringBuilder queryBuilder =
+			new StringBuilder(FIND_GERMPLASM_MATCHES_MAIN_QUERY);
```
darla-leafnode, line 1721: I think we can reduce duplicated here if we just add a match by GID in `GermplasmMatchRequestDto` 
`private List<String> gids;`
then expand `addGermplasmMatchesFilter` to filter by gids. Then getGermplasmDtoByGid here can just call `findGermplasmMatches` returning the first DTO matched by GID (if any)

src/main/java/org/generationcp/middleware/dao/GermplasmStudySourceDAO.java
```diff
@@ -119,6 +166,61 @@ public long countGermplasmStudySourceList(final GermplasmStudySourceSearchReques
...
+			final List<Map<String, Object>> results = sqlQuery.list();
+
+			if (!results.isEmpty()) {
+				final Map<String, Object> result = results.get(0);
+				final GermplasmOriginDto germplasmOriginDto = new GermplasmOriginDto();
+				germplasmOriginDto.setStudyId((Integer) result.get("studyId"));
+				germplasmOriginDto.setStudyName((String) result.get("studyName"));
+				germplasmOriginDto.setObservationUnitId((String) result.get("observationUnitId"));
+				germplasmOriginDto.setPlotNumber((Integer) result.get("plotNumber"));
+				germplasmOriginDto.setBlockNumber(((result.get("blockNumber") != null)) ? (Integer) result.get("plotNumber") : null);
```
darla-leafnode, line 193: We should be setting `blockNumber` not `plotNumber`?

src/main/java/org/generationcp/middleware/dao/ProgenitorDAO.java
```diff
@@ -44,17 +45,16 @@ public Progenitor getByGIDAndPID(final Integer gid, final Integer pid) {
...
-			}
+
+			Criteria criteria = this.getSession().createCriteria(this.getPersistentClass());
+			criteria.createAlias("germplasm", "germplasm");
+			criteria.add(Restrictions.eq("germplasm.gid", gid));
+			return criteria.list();
+
 		} catch (HibernateException e) {
 			throw new MiddlewareQueryException(
 				"Error with getByGIDAndPID(gid=" + gid + ") query from Progenitor: " + e.getMessage(), e);
```
darla-leafnode, line 56: Existing in the master but should be "Error with **getByGID**(gid="

src/main/java/org/generationcp/middleware/domain/germplasm/GermplasmDto.java
```diff
@@ -32,114 +32,154 @@
...
-		return breedingMethodId;
+		return this.breedingMethodId;
 	}
 
 	public void setBreedingMethodId(final Integer breedingMethodId) {
 		this.breedingMethodId = breedingMethodId;
 	}
 
+	public Integer getGpid1() {
+		return gpid1;
```
darla-leafnode, line 152: return this.gpid1

src/main/java/org/generationcp/middleware/domain/germplasm/GermplasmDto.java
```diff
@@ -32,114 +32,154 @@
...
+	}
+
+	public void setGpid1(final Integer gpid1) {
+		this.gpid1 = gpid1;
+	}
+
+	public Integer getGpid2() {
+		return gpid2;
```
darla-leafnode, line 160: return this.gpid2

src/main/java/org/generationcp/middleware/domain/germplasm/GermplasmDto.java
```diff
@@ -32,114 +32,154 @@
...
+	}
+
+	public void setGpid2(final Integer gpid2) {
+		this.gpid2 = gpid2;
+	}
+
+	public List<Integer> getOtherProgenitors() {
+		return otherProgenitors;
```
darla-leafnode, line 168: return this.otherProgenitors

src/main/java/org/generationcp/middleware/domain/germplasm/GermplasmNameDto.java
```diff
@@ -78,6 +82,22 @@ public void setGid(final Integer gid) {
 		this.gid = gid;
 	}
 
+	public Integer getId() {
+		return id;
```
darla-leafnode, line 86: return this.id

src/main/java/org/generationcp/middleware/domain/germplasm/GermplasmNameDto.java
```diff
@@ -78,6 +82,22 @@ public void setGid(final Integer gid) {
...
+	}
+
+	public void setId(final Integer id) {
+		this.id = id;
+	}
+
+	public boolean isPreferred() {
+		return preferred;
```
darla-leafnode, line 94: return this.preferred

src/main/java/org/generationcp/middleware/domain/germplasm/ProgenitorsDetailsDto.java
```diff
@@ -0,0 +1,101 @@
...
+	private Integer breedingMethodId;
+	private String breedingMethodName;
+	private String breedingMethodCode;
+	private String breedingMethodType;
+	private GermplasmDto femaleParent;
+	private List<GermplasmDto> maleParents;
+	private GermplasmDto groupSource;
+	private GermplasmDto immediateSource;
+
+	public Integer getBreedingMethodId() {
```
darla-leafnode, line 22: Run cleanup here

src/main/java/org/generationcp/middleware/service/impl/study/SampleServiceImpl.java
```diff
@@ -225,10 +215,17 @@ public Boolean studyHasSamples(final Integer studyId) {
...
 	}
 
+	void populateTakenBy(final List<SampleDTO> sampleDTOS) {
+		// Populate takenBy with full name of user from workbench database.
+		final List<Integer> userIds = sampleDTOS.stream().map(sampleDTO -> sampleDTO.getTakenByUserId()).collect(Collectors.toList());
+		final Map<Integer, String> userIDFullNameMap = this.userService.getUserIDFullNameMap(userIds);
+		sampleDTOS.forEach(sampleDTO -> sampleDTO.setTakenBy(userIDFullNameMap.get(sampleDTO.getTakenByUserId())));
+	}
+	
 	protected void setWorkbenchDataManager(final WorkbenchDataManager workbenchDataManager) {
```
darla-leafnode, line 229: Existing in master but seems we can remove `workbenchDataManager` from this class

src/main/java/org/generationcp/middleware/dao/GermplasmDAO.java
clarysabel: @darla-leafnode  I wont mess with the matches query, It was complicated to find an index that works well by the time I implemented it. That's the main reason I have created another query, so I am sure I dont affect the matches performance that it is used in the new import 
CC @abatac

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmAttributeServiceImpl.java
clarysabel: According to the mockup it is actually possible to modify the date, It seems it is not a system date but a date set by breeders

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmService.java
clarysabel: We have the same problem with attributes but we know there should be a refactor. As of now I would not mind that much on where we locate this function. It was previously in BMSAPI calling 
`germplasmService.getUserDefinedFieldByTableTypeAndCodes `
which was also weird.
I prefer to wait a bit to check what stays as part of user defined fields and what will be moved to ontology to make a decision on where we should place all this kind of metadata code

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmServiceImpl.java
clarysabel: that was me... I found that having constructors with a bunch of parameters is hard to read and unuseful when most of them are not mandatory. If have to chose I will prefer a Builder than having what we have now, creating a bunch of constructor and then we end passing null

PR #1010, opened 2021-03-26
IBP-4358-GidWithProgenyByMID

gelli-leafnode:
Newly added query, also filters progeny's method type. 
Modified methods to accept method type id list (List of String)


gelli-leafnode:
Code preview is not updated.

PR #1021, opened 2021-04-15
IBP-4476-ChangesetRenameVariableXIfExists

gelli-leafnode:
New Changesets for checking if variable 'X' exists, if true rename to 'X (old)' - v17.4.0-01
This changeset should run before the changeset_id v17.4.0-1


clarysabel (changes requested):
Rejected because it can create data issues in phenotype. See my comment in the ticket

PR #1027, opened 2021-04-21
Ibp 4307 inventory manager search performance

lik-83:


nahuel-soldevilla (approved):
Looks good. As we discussed, transactions is still not a problem, so we can wait and/or tackle in another ticket

PR #1038, opened 2021-04-28
IBP-4542 Germplasm Progenitor Update in Import Germplasm Updates

abatac:


clarysabel:
@abatac Glad you reuse the functions so we have all this logic in only one plase

PR #1045, opened 2021-05-03
IBP-4341: added changeset to remove germplasm update permission to re…

lik-83:
…adonly role

### DIFF
--- a/src/main/resources/liquibase/workbench_changelog/17_4_1.xml
+++ b/src/main/resources/liquibase/workbench_changelog/17_4_1.xml
@@ -35,7 +35,7 @@
 				FROM role_permission rp
 						 INNER JOIN role r ON rp.role_id = r.id
 						 INNER JOIN permission p ON rp.permission_id = p.permission_id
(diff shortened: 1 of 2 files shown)


src/main/resources/liquibase/workbench_changelog/17_4_1.xml
```diff
@@ -28,5 +28,25 @@
...
 
-</databaseChangeLog>
+	<changeSet author="lkovacic" id="v17.4.1-3">
+		<preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="4">
+                SELECT count(*)
+				FROM role_permission rp
+						 INNER JOIN role r ON rp.role_id = r.id
+						 INNER JOIN permission p ON rp.permission_id = p.permission_id
+				WHERE r.name = 'Readonly' AND p.name IN ('IMPORT_GERMPLASM', 'SEARCH_GERMPLASM', 'IMPORT_GERMPLASM_UPDATES', 'GERMPLASM_LABEL_PRINTING');
```
clarysabel, line 38: I think we should use role id = 4 instead of the name (or use both), because the role name is modifiable. 
What do you think @nahuel-soldevilla

lik-83: Done!

PR #1047, opened 2021-05-05
IBP-4406 Fix BRAVA Errors in v2 APIs

abatac:


src/main/java/org/generationcp/middleware/dao/LocationDAO.java
```diff
@@ -1014,10 +1015,22 @@ public long countLocations(final LocationSearchRequest locationSearchRequest) {
...
+			Coordinate coordinate = null;
+			final Double longitude = (Double) result.get("longitude");
+			final Double latitude = (Double) result.get("latitude");
+			final Double altitude = (Double) result.get("altitude");
+			if (longitude != null && latitude != null) {
+				final List<Double> coordinatesList = new ArrayList<>();
+				coordinatesList.add(longitude);
+				coordinatesList.add(latitude);
+				if (altitude != null) {
+					// Only add altitube if available.
```
darla-leafnode, line 1028: typo error, should be "altitude"

src/main/java/org/generationcp/middleware/util/serializer/DatePropertySerializer.java
```diff
@@ -0,0 +1,42 @@
...
+	@Override
+	public void serialize(final Date value, final JsonGenerator gen, final SerializerProvider serializers)
+		throws IOException {
+
+		if (value == null) {
+			gen.writeNull();
+			return;
+		}
+
+		if (serializers.getActiveView() != null) {
```
darla-leafnode, line 29: Can be simplified as:
`if (serializers.getActiveView() != null && BrapiView.BrapiV2.class.equals(serializers.getActiveView())) {
gen.writeString(this.formatterWithTime.format(value));
} else {
gen.writeString(this.formatterSimple.format(value));
}
`

darla-leafnode: End date is nullable. How is it being handled/displayed now?

src/main/java/org/generationcp/middleware/util/serializer/SynonymPropertySerializer.java
```diff
@@ -0,0 +1,33 @@
...
+		throws IOException {
+
+		if (serializers.getActiveView() != null && CollectionUtils.isNotEmpty(value)) {
+			if (BrapiView.BrapiV1_3.class.equals(serializers.getActiveView()) || BrapiView.BrapiV1_2.class
+				.equals(serializers.getActiveView())) {
+				gen.writeObject(value.stream().map(Synonym::getSynonym).collect(Collectors.toList()));
+			} else {
+				gen.writeObject(value);
+			}
+		} else {
```
darla-leafnode, line 27: Why do we need this final else clause here (ie. when there are no synonyms)?

src/main/java/org/generationcp/middleware/util/serializer/SeasonPropertySerializer.java
```diff
@@ -0,0 +1,31 @@
...
+	public void serialize(final List<SeasonDto> value, final JsonGenerator gen, final SerializerProvider serializers)
+		throws IOException {
+
+		if (serializers.getActiveView() != null && CollectionUtils.isNotEmpty(value)) {
+			if (BrapiView.BrapiV2.class.equals(serializers.getActiveView())) {
+				gen.writeObject(value.stream().map(SeasonDto::getSeason).collect(Collectors.toList()));
+			} else {
+				gen.writeObject(value);
+			}
+		} else {
```
darla-leafnode, line 26: Same here, why do we need to write the object when there is no season list/value?

src/main/java/org/generationcp/middleware/dao/LocationDAO.java
abatac: Done

src/main/java/org/generationcp/middleware/util/serializer/DatePropertySerializer.java
abatac: Done. If end date is null, it will not be included in json response.

src/main/java/org/generationcp/middleware/util/serializer/SynonymPropertySerializer.java
abatac: If synonym is empty, we need to write/return it as is, if not, JsonSerializer will return an error "Can not write a field name, expecting a value"

src/main/java/org/generationcp/middleware/util/serializer/SeasonPropertySerializer.java
abatac: Same with synonym.

darla-leafnode (approved):

PR #1051, opened 2021-05-11
IBP-4489 Location and Method Dropdown Code Clean Up

abatac:


src/main/java/org/generationcp/middleware/api/breedingmethod/BreedingMethodDTO.java
```diff
@@ -15,6 +15,7 @@
 	private Integer methodClass;
 	private Integer mid;
 	private Integer numberOfProgenitors;
+	private Boolean bulkingMethod;
```
darla-leafnode, line 18: Rename as `isBulkingMethod` per standards

abatac: Done

darla-leafnode (approved):

PR #1055, opened 2021-05-19
Ibp 4525 germplasm audit

lik-83:


src/main/resources/liquibase/crop_changelog/18_1_0.xml
```diff
@@ -214,4 +214,244 @@
...
+		<comment>
+            Add data to 'audit_cfg' table
+        </comment>
+        <sql dbms="mysql" splitStatements="true">
+			INSERT INTO audit_cfg (entity_name, is_audited)
+			VALUES ('GERMPLSM', 1),
+				('ATRIBUTS', 1),
+				('BIBREFS', 1),
+				('NAMES', 1),
+				('EXTERNAL_REFERENCE', 1),
```
clarysabel, line 251: I think we can just rename it to GERMPLASM_EXTERNAL_REFERENCE now so we dont affect this row when renaming the table @lik-83

PR #1070, opened 2021-06-14
IBP-4633 Group and Ungroup Germplasm API

darla-leafnode:


src/main/java/org/generationcp/middleware/service/impl/GermplasmGroupingServiceImpl.java
```diff
@@ -43,76 +46,66 @@
...
+	}
+
+	private List<GermplasmGroup> getGroupMembersForGermplasm(final List<Germplasm> germplasmList) {
+		final List<GermplasmGroup> germplasmGroupList = new ArrayList<>();
+		for (final Germplasm founder : germplasmList) {
+			final GermplasmGroup germplasmGroup = new GermplasmGroup();
 
-		final Method method = this.methodDAO.getById(founder.getMethodId());
-		founder.setMethod(method);
+			final Method method = this.daoFactory.getMethodDAO().getById(founder.getMethodId());
```
abatac, line 98: I think method is already available in germplasm.getMethod(). No need to get it from DAO.

PR #1071, opened 2021-06-15
IBP-4634 Move Coding to API

abatac:


src/main/java/org/generationcp/middleware/domain/germplasm/GermplasmCodeNameBatchRequestDto.java
```diff
@@ -0,0 +1,54 @@
+package org.generationcp.middleware.domain.germplasm;
+
+import org.generationcp.middleware.pojos.germplasm.GermplasmNameSetting;
+import org.pojomatic.Pojomatic;
+import org.pojomatic.annotations.AutoProperty;
+
+import java.util.List;
+
+@AutoProperty
+public class GermplasmCodeNameBatchRequestDto {
```
darla-leafnode, line 10: Is this DTO specific to germplasm code name or can be it a generic germplasm name batch request dto?

abatac: Yep this is specific to germplasm code name because we need to pass the germplasm name settings, and it only supports CODE1,2,3. Other names don't have that option.

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmService.java
```diff
@@ -28,6 +28,10 @@
 
 public interface GermplasmService {
 
+	Germplasm getGermplasmByGID(Integer gid);
```
darla-leafnode, line 31: As much as possible we shouldn't be exposing Germplasm object - the raw Hibernate entity from MW. It seems we are interested in the mgid property so maybe we can just return a map of mgid for all GIDS to code
`Map<Integer, Integer> getGermplasmGroupIdMap`
with gid as key and mgid as map value.

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmService.java
```diff
@@ -28,6 +28,10 @@
...
+
+	Germplasm saveGermplasm(Germplasm germplasm);
```
darla-leafnode, line 33: Same here as much as possible let's not expose Germplasm entity, especially for saving. Since we are saving germplasm names, I think we should use GermplasmNamingService for saving names

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmService.java
clarysabel: I think we should stop returning Maps as much as we can because these functions are not reusable at all. Maps should be created where they are really needed. 
For what he needs he can use GermplasmDto and we have some functions in place returning this object

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmNameServiceImpl.java
```diff
@@ -108,4 +111,9 @@ public Integer createName(final GermplasmNameRequestDto germplasmNameRequestDto,
 		daoFactory.getNameDao().save(name);
 		return name.getNid();
 	}
+
+	@Override
+	public List<GermplasmNameDto> getGermplasmNamesByGids(final List<Integer> gids) {
+		return this.daoFactory.getNameDao().getGermplasmNamesByGids(new ArrayList<>(gids));
```
darla-leafnode, line 117: Why instantiate a new array list for the gids parameter?

abatac: I changed the gids param to Set type before so I had to convert, just forgot to correct this.

darla-leafnode (approved):

PR #1077, opened 2021-06-18
IBP-4704 include passport in filter by attributes

mderamos:


darla-leafnode:
@clarysabel -   what do you think about merging this for 18.2? I think this code will be obsolete soon when we merge then attributes refactoring

clarysabel:
HI @darla-leafnode 
It does not make sense to merge it. 
It was solved in IBP-4440 and it is retrieving information from cvterm now.

PR #1083, opened 2021-06-29
Ibp 4746 germplasm details audit changes 2

lik-83:


src/main/resources/liquibase/workbench_changelog/18_3_0.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+	<changeSet author="lkovacic" id="v18.3.0-1">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">
+                SELECT count(*) FROM permission where name = 'VIEW_CHANGE_HISTORY'
```
clarysabel, line 10: I will rename this to VIEW_GERMPLASM_CHANGE_HISTORY
@lik-83

src/main/java/org/generationcp/middleware/dao/audit/germplasm/GermplasmBasicDetailsAuditDAOQuery.java
```diff
@@ -0,0 +1,48 @@
...
+
+class GermplasmBasicDetailsAuditDAOQuery {
+
+	static final String LOCATION_NAME_ALIAS = "locationName";
+	static final String CREATION_DATE_ALIAS = "creationDate";
+
+	static final String LOCATION_CHANGED_ALIAS = "locationChanged";
+	static final String CREATION_DATE_CHANGED_ALIAS = "creationDateChanged";
+
+	private final static String BASE_QUERY = "SELECT "
```
clarysabel, line 13: not sure how we will handle it in the future but I think we should use LEFT JOIN instead of INNER JOIN for the cases where the references where deleted. The problem with location and breeding methods is that we are showing the names and not the ids, so if the reference was deleted I dont know what we will show. We can leave it as it is now and continue the discussion and see how we can solve it 
What do you think @lik-83

src/main/resources/liquibase/workbench_changelog/18_3_0.xml
```diff
@@ -0,0 +1,27 @@
...
+            <sqlCheck expectedResult="0">
+                SELECT count(*) FROM permission where name = 'VIEW_GERMPLASM_CHANGE_HISTORY'
+            </sqlCheck>
+        </preConditions>
+        <comment>Add Manage Germplasm "View Germplasm change history" permission</comment>
+		<sql dbms="mysql" splitStatements="true">
+			set @manage_germplasm_id = (select permission_id from permission where name = 'MANAGE_GERMPLASM');
+
+			insert into permission(name, description, parent_id, workbench_sidebar_category_link_id)
+			values ('VIEW_GERMPLASM_CHANGE_HISTORY', 'View Germplasm change history', @manage_germplasm_id, null);
```
clarysabel, line 18: Remove Germplasm from the description: View Germplasm change history
@lik-83

PR #1087, opened 2021-07-05
Replace usage of INSERT INTO by hibernate in fieldmap

clarysabel:
IBP-4431


src/main/java/org/generationcp/middleware/operation/saver/ExperimentPropertySaver.java
```diff
@@ -31,11 +30,11 @@
 
 public class ExperimentPropertySaver extends Saver {
 
-	private StringBuilder batchExperimentPropInsertSql = new StringBuilder();
+	private final StringBuilder batchExperimentPropInsertSql = new StringBuilder();
```
nahuel-soldevilla, line 33: we can remove this field and probably rename `saveInBatchOrUpdateProperty` to just `saveOrUpdateProperty`

nahuel-soldevilla (approved):

PR #1093, opened 2021-07-14
IBP-4724 fix samples brapi brava validator errors

mderamos:


src/main/java/org/generationcp/middleware/service/impl/study/SampleServiceImpl.java
```diff
@@ -147,6 +147,8 @@ private SampleDetailsDTO getSampleDetailsDTO(final Sample sample) {
 		samplesDetailsDto.setEntryNo(Integer.valueOf(entryNo));
```
abatac, line 147: Remove unused class variable: workbenchDataManager.

abatac (changes requested):

PR #1094, opened 2021-07-14
IBP-4732 ksu fieldbook image integration

nahuel-soldevilla:
- file_metadata table: store some info (name, path, file_uuid, fk to nd_exp) from files before uploading to file storage
- file metadata services
- UIDGenerator refactor

TODO

- [ ] save file metadata coming from manage studies file upload


src/main/java/org/generationcp/middleware/pojos/file/FileMetadata.java
```diff
@@ -0,0 +1,159 @@
+package org.generationcp.middleware.pojos.file;
+
+import org.generationcp.middleware.pojos.AbstractEntity;
+import org.generationcp.middleware.pojos.dms.ExperimentModel;
+
+import javax.persistence.*;
```
clarysabel, line 6: with our settings, I think we should not be importing the whole package, right? @nahuel-soldevilla 
Implementing equals, hashCode and toString is missing in this and another classes. 
Does it worth to at least implement it in the entity?

src/main/java/org/generationcp/middleware/util/uid/UIDGenerator.java
```diff
@@ -0,0 +1,70 @@
...
+import org.generationcp.middleware.pojos.workbench.CropType;
+import org.springframework.util.CollectionUtils;
+
+import java.util.List;
+import java.util.UUID;
+
+public final class UIDGenerator {
+
+	public static enum UID_ROOT {
+		LOT("L"),
```
clarysabel, line 15: dont we have same generation strategy for samples? 
was this change tested as a regression?

src/main/java/org/generationcp/middleware/pojos/file/FileMetadata.java
nahuel-soldevilla: >with our settings, I think we should not be importing the whole package, right?
Right. I created the entity from the table using intellij plugin (POJO generator), and it generates it that way.
I cleaned up the imports and added toString.
I'm not sure if it's worth to add hashCode and equals, perhaps we can wait for a scenario where we actually needed

src/main/java/org/generationcp/middleware/util/uid/UIDGenerator.java
nahuel-soldevilla: As we discussed, I tested the generation for observation units, lots and germplasm, and logged the task to implement for samples in https://ibplatform.atlassian.net/browse/IBP-4167

src/main/java/org/generationcp/middleware/pojos/file/FileMetadata.java
clarysabel: :+1:

PR #1100, opened 2021-07-21
IBP-4732 Remove crop name from table creation changeset

darla-leafnode:


darla-leafnode:
@nahuel-soldevilla , @clarysabel - we modified the previous changeset to remove db name

abatac (approved):

PR #1104, opened 2021-07-29
IBP-4853 fix brapi call issues

abatac:


src/main/java/org/generationcp/middleware/service/impl/study/StudyInstanceServiceImpl.java
```diff
@@ -337,78 +334,72 @@ public InstanceDescriptorData addInstanceDescriptorData(final InstanceDescriptor
...
 			}
-			return null;
-		} catch (final MiddlewareQueryException e) {
-			final String message = "Error with getStudyDetailsForGeolocation() query with instanceId: " + instanceId;
-			StudyInstanceServiceImpl.LOG.error(message, e);
-			throw new MiddlewareQueryException(message, e);
+			studyDetailsDto.setAdditionalInfo(properties);
+			return studyDetailsDto;
 		}
+		return null;
```
darla-leafnode, line 401: Returning null should be used very sparingly. Can we try to convert this API to:
Optional<StudyDetailsDto> getStudyDetailsByInstance(final Integer instanceId)

darla-leafnode (approved):

PR #1107, opened 2021-08-03
Ibp 4455 fix germplasm origin schema

abatac:


darla-leafnode:
@abatac , in this case, we will not need the constant `ORIGIN_ATTR` in `GermplasmImportRequest`. Please delete it and remove its usages in this class and any test class

darla-leafnode (approved):

PR #1114, opened 2021-08-19
IBP-4752: added service to get studies filtered by name

lik-83:


nahuel-soldevilla (approved):
Looks good. 

For context, what we discussed: 

- The option of expanding the pageSize in the frontend was a hack, so we went for the option of pagination
- Since normally in brapi endpoints we consider exact match for filter params, we switched the frontend to use an internal api (the graphical queries header is not strictly part of the brapp, so no issue in using internal api)
- We start with a "simple" implementation for /studies and expand later if needed

PR #1117, opened 2021-08-20
IBP-4822 browse study error when enviroments have the same location name

cuenyad:
Please, review.

Included:
https://github.com/IntegratedBreedingPlatform/Commons/pull/320
https://github.com/IntegratedBreedingPlatform/Fieldbook/pull/1025

### DIFF
--- a/src/main/java/org/generationcp/middleware/manager/StudyDataManagerImpl.java
+++ b/src/main/java/org/generationcp/middleware/manager/StudyDataManagerImpl.java
@@ -1108,9 +1108,9 @@ public class StudyDataManagerImpl extends DataManager implements StudyDataManage
 	}
 
 	@Override
-	public BiMap<String, String> createInstanceLocationIdToNameMapFromStudy(final int studyId) {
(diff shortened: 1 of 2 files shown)


nahuel-soldevilla (approved):
@cuenyad This looks fair to me as a workaround, as we've been discussing.

For reference, and please correct if I'm wrong, the two problems we have in the code caused by the fact that we have locations with the same name:

- BiMap is not suitable as it requires uniqueness of values and keys, however for most scenarios a HashMap would work because we only need to retrieve the location name based on its id.
- The particular case of BreedingView import, where the inverse problem happens, and it's not possible to know the id of the location based on the location name when no other information (e.g instance number) is available. For this case, as we discussed, we'll show a proper error message and create an improvement to redesign that workflow.

cuenyad:
Hi @nahuel-soldevilla regarding the BreedingView import I created the [IBP-4951](https://ibplatform.atlassian.net/browse/IBP-4951) to discuss with Mariano and fix it.

darla-leafnode (approved):

PR #1119, opened 2021-08-23
IBP-4917: Associate files only to dataset variables

nahuel-soldevilla:
BMSAPI:
- new /variables/filter param: datasetIds
Middleware:
- OntologyVariableDataManagerImpl.getWithFilter is candidate for
refactor but the combination of filters (datasetId+variableTypeIds)
 works well for this particular scenario, while other approaches don't
 (variableIds+variableTypeIds, supposing we would want pre-gather the
 list of variableIds for the dataset)


darla-leafnode:
Please apply code formatter and cleanup on affected classes

darla-leafnode (approved):

PR #1132, opened 2021-09-02
IBP-4878 Remove AppLauncherService

cuenyad:
As a part of removing the AppLauncherService implementation.

* Removed unnecessary properties.
* Removed unnecessary tests.
* Removed methods from WorkbenchDataManager without use.

Included:
https://github.com/IntegratedBreedingPlatform/BMSAPI/pull/582

Please, review

### DIFF
--- a/src/test/java/org/generationcp/middleware/dao/ProjectDAOTest.java
+++ b/src/test/java/org/generationcp/middleware/dao/ProjectDAOTest.java
@@ -136,36 +136,6 @@ public class ProjectDAOTest extends IntegrationTestBase {
 	}
 
 	@Test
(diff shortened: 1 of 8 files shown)


src/test/java/org/generationcp/middleware/dao/ProjectDAOTest.java
```diff
@@ -135,36 +135,6 @@ public void setup() {
 
 	}
 
-	@Test
```
clarysabel, line 138: why the tests that are about project are removed?
@cuenyad

cuenyad: Reverted change and fixied test.

PR #1137, opened 2021-09-14
IBP-4987: added validation for deleted germplasms on import dataset

cheryleigh:
- user should not be able to create a new study using the Import Dataset feature with deleted germplasm (show error same with non-existing gid)


mderamos (approved):

src/main/java/org/generationcp/middleware/pojos/Germplasm.java
```diff
@@ -62,7 +62,7 @@
 	@NamedQuery(name = "getAllGermplasm", query = "FROM Germplasm"),
 	@NamedQuery(name = "countAllGermplasm", query = "SELECT COUNT(g) FROM Germplasm g"),
 
-	@NamedQuery(name = "countMatchGermplasmInList", query = "SELECT COUNT(g) FROM Germplasm g WHERE g.gid IN (:gids)")
+	@NamedQuery(name = "countMatchGermplasmInList", query = "SELECT COUNT(g) FROM Germplasm g WHERE g.gid IN (:gids) AND g.deleted = 0")
```
mderamos, line 65: Hi @cheryleigh 
The call for this query is called in GermplasmDataManagerImpl. Move it to GermplasmService and call it from there. We're slowly deprecating Data Managers and moving the methods to the Service classes.

src/main/java/org/generationcp/middleware/api/germplasm/GermplasmService.java
```diff
@@ -55,6 +55,14 @@
 
 	List<GermplasmDto> findGermplasmMatches(GermplasmMatchRequestDto germplasmMatchRequestDto, Pageable pageable);
 
+	/**
+	 * Returns the count of records that matched the gids in the specified list.
+	 *
+	 * @param gids - The list of gids to match for in the germplasm table
+	 * @return
+	 */
+	long countMatchGermplasmInList(Set<Integer> gids);
```
darla-leafnode, line 64: Hi @cheryleigh - please explore if you can use/reuse `GermplasmService#countSearchGermplasm` instead and remove all code related to `countMatchGermplasmInList` - including the query - altogether

src/main/java/org/generationcp/middleware/service/api/DataImportService.java
```diff
@@ -158,7 +158,7 @@ int saveProjectData(final Workbook workbook, final String programUUID, final Cro
 
 	Optional<MeasurementVariable> findMeasurementVariableByTermId(int termId, List<MeasurementVariable> list);
 
-	void checkForInvalidGids(final Workbook workbook, final List<Message> messages);
+	void checkForInvalidGids(final Workbook workbook, final List<Message> messages, final String programUUID);
```
darla-leafnode, line 161: Remove final from API parameters (apply to all similar cases in this class)

darla-leafnode (approved):

PR #1139, opened 2021-09-16
IBP-5017 remove program UUID from methods

nahuel-soldevilla:


src/main/java/org/generationcp/middleware/api/breedingmethod/BreedingMethodSearchRequest.java
```diff
@@ -9,6 +9,9 @@
 @AutoProperty
 public class BreedingMethodSearchRequest {
 
+	/**
+	 * Used to get program favorites
+	 */
```
clarysabel, line 14: @nahuel-soldevilla Maybe we can remove the favoritesOnly? If the programUUID is set then it menas it will only retrieve favorites for that program 
It can be done in another ticket but seems it does not make sense to have 2 fields for that

clarysabel (approved):

PR #1144, opened 2021-09-29
IBP-4973 GermplasmGroupingServiceImpl optimization

cheryleigh:
Changes:
- use germplasm.getMethod() instead of getMethodDao
- process selection history name type only once (outside the gid loop)
- store germplasm selection history name per germplasm in a map to avoid multiple retrieval process
- remove maintenance children retrieval in GermplasmDAO#getAllChildren

### DIFF
--- a/src/main/java/org/generationcp/middleware/service/impl/GermplasmGroupingServiceImpl.java
+++ b/src/main/java/org/generationcp/middleware/service/impl/GermplasmGroupingServiceImpl.java
@@ -53,6 +53,10 @@ public class GermplasmGroupingServiceImpl implements GermplasmGroupingService {
 	@Value("${germplasm.grouping.max.recursion}")
 	public int maxRecursiveQueries;
 
+	private UserDefinedField selHistNameType;
+
+	private UserDefinedField selHistAtFixationNameType;
+
 	private DaoFactory daoFactory;
 
 	public GermplasmGroupingServiceImpl(final HibernateSessionProvider sessionProvider) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/generationcp/middleware/service/impl/GermplasmGroupingServiceImpl.java
```diff
@@ -53,8 +55,14 @@
 	@Value("${germplasm.grouping.max.recursion}")
 	public int maxRecursiveQueries;
 
+	private UserDefinedField selHistNameType;
+
+	private UserDefinedField selHistAtFixationNameType;
+
 	private DaoFactory daoFactory;
 
+	private Map<Integer, String> germplasmSelHistNameMap = new HashMap<>();
```
darla-leafnode, line 64: declare with final

src/main/java/org/generationcp/middleware/service/impl/GermplasmGroupingServiceImpl.java
```diff
@@ -64,6 +72,9 @@ public GermplasmGroupingServiceImpl(final HibernateSessionProvider sessionProvid
 		final List<Germplasm> germplasmList = this.daoFactory.getGermplasmDao().getByGIDList(gids);
 		final Map<Integer, Germplasm> germplasmMap = germplasmList.stream().collect(Collectors.toMap(
 			Germplasm::getGid, Function.identity()));
+		this.selHistNameType = getSelectionHistoryNameType(SELECTION_HISTORY_NAME_CODE);
```
darla-leafnode, line 75: run code cleanup

darla-leafnode (approved):

PR #1175, opened 2021-11-25
IBP-4793 can not edit or delete system variables

cuenyad:
Included:
https://github.com/IntegratedBreedingPlatform/BMSAPI/pull/624
https://github.com/IntegratedBreedingPlatform/Workbench/pull/591

Please, review.

### DIFF
--- a/src/main/resources/liquibase/crop_changelog/19_4_0.xml
+++ b/src/main/resources/liquibase/crop_changelog/19_4_0.xml
@@ -10,7 +10,7 @@
 				<columnExists tableName="cvterm" columnName="is_system"/>
 			</not>
 		</preConditions>
-		<comment>Adding is_system_term column</comment>
+		<comment>Adding is_system column</comment>
         <sql>
 			ALTER TABLE cvterm ADD COLUMN is_system INT(11) NOT NULL DEFAULT '0' AFTER is_relationshiptype;
         </sql>
--- a/src/main/resources/liquibase/crop_changelog/19_4_0.xml
+++ b/src/main/resources/liquibase/crop_changelog/19_4_0.xml
@@ -23,12 +23,12 @@
 		<comment>Mark system variables</comment>
         <sql>
 			update cvterm cv INNER JOIN cvterm cv1 ON cv.cvterm_id = cv1.cvterm_id  set cv.is_system = 1
-			where cv1.name in ('PLOT_NUMBER_AP_text','INSTANCE_NUMBER_AP_text','REP_NUMBER_AP_text','PLANT_NUMBER_AP_text','PLOTCODE_AP_text');
-
-			update cvterm cv INNER JOIN cvterm cv1 ON cv.cvterm_id = cv1.cvterm_id  set cv.is_system = 1
-			where cv1.name in ('TRIAL_INSTANCE','LOCATION_ID','EXPT_DESIGN','FIELDMAP COLUMN'
-				,'FIELDMAP RANGE','OBS_UNIT_ID','COL','ROW','BLOCK_NO','PLOT_NO','REP_NO'
-				,'ENTRY_CODE','ENTRY_NO','DESIGNATION','GID','ENTRY_TYPE', 'PLOT_CODE', 'CHECK_INTERVAL', 'CHECK_PLAN', 'CHECK_START');
+			where cv1.name in ('TRIAL_INSTANCE','LOCATION_ID','LOCATION_NAME','EXPT_DESIGN','FIELDMAP COLUMN'
+				,'FIELDMAP RANGE','OBS_UNIT_ID','COL','ROW','BLOCK_NO','PLOT_NO','REP_NO','ENTRY_CODE'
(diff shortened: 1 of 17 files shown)


src/main/resources/liquibase/crop_changelog/19_4_0.xml
```diff
@@ -0,0 +1,34 @@
...
+    </changeSet>
+
+	<changeSet author="cuenyad" id="v19.4.0-2">
+		<preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="1">SELECT count(1) > 0 FROM cvterm where cvterm_id &lt;= 100000 and cv_id != 1040 and is_system = 0</sqlCheck>
+		</preConditions>
+		<comment>Mark system variables</comment>
+        <sql>
+			update cvterm cv INNER JOIN cvterm cv1 ON cv.cvterm_id = cv1.cvterm_id  set cv.is_system = 1
+			where cv1.name in ('TRIAL_INSTANCE','LOCATION_ID','LOCATION_NAME','EXPT_DESIGN','FIELDMAP COLUMN'
```
nahuel-soldevilla, line 26: Another possibility, given that it could be difficult to find just one query that matches all of what we want to mark as _system variable_, is to use a generic query to mark the _majority_ of the terms we want, and then use a separate update to mark the _edge cases_ that didn't fall in the first criteria

nahuel-soldevilla (approved):

src/main/resources/liquibase/crop_changelog/19_4_0.xml
```diff
@@ -0,0 +1,40 @@
...
+
+	<changeSet author="cuenyad" id="v19.4.0-1">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<columnExists tableName="cvterm" columnName="is_system"/>
+			</not>
+		</preConditions>
+		<comment>Adding is_system column</comment>
+        <sql>
+			ALTER TABLE cvterm ADD COLUMN is_system INT(11) NOT NULL DEFAULT '0' AFTER is_relationshiptype;
```
clarysabel, line 15: seems that `tinyint(1) unsigned zerofill` fits better to the is_system

src/main/resources/liquibase/crop_changelog/19_4_0.xml
```diff
@@ -0,0 +1,40 @@
...
+        </sql>
+    </changeSet>
+
+	<changeSet author="cuenyad" id="v19.4.0-2">
+		<preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="1">SELECT count(1) > 0 FROM cvterm where is_system = 0 and name in ('TRIAL_INSTANCE','LOCATION_ID','LOCATION_NAME','EXPT_DESIGN','FIELDMAP COLUMN'
```
clarysabel, line 21: I think that this precondition can be simplified, if is_system exists, then the variables can be marked as is_system. I dont think it is necessary to check if the value is 0

src/main/resources/liquibase/crop_changelog/19_4_0.xml
```diff
@@ -0,0 +1,40 @@
...
+			    ,'PLANT_NUMBER_AP_text','PLOTCODE_AP_text');
+			</sqlCheck>
+		</preConditions>
+		<comment>Mark system variables</comment>
+        <sql>
+			update cvterm cv INNER JOIN cvterm cv1 ON cv.cvterm_id = cv1.cvterm_id  set cv.is_system = 1
+			where cv1.name in ('TRIAL_INSTANCE','LOCATION_ID','LOCATION_NAME','EXPT_DESIGN','FIELDMAP COLUMN'
+				,'FIELDMAP RANGE','OBS_UNIT_ID','COL','ROW','BLOCK_NO','PLOT_NO','REP_NO','ENTRY_CODE'
+			    ,'ENTRY_NO','DESIGNATION','GID','MGID','FGID','GROUPGID','EXPT_DESIGN_SOURCE','DATASET_NAME'
+			    ,'DATASET_TITLE','LOCATION_NAME','DESIG','CROSS','ENTRY_TYPE', 'PLOT_CODE', 'CHECK_INTERVAL'
```
clarysabel, line 35: For the ones that are used in TermId, I suggest to identify them by cvterm_id instead than name. 

The uniqueness of the cvterm table is not given by the name, so if there is more than one with the same name and different type, they will also be marked as system variables. Add cv_id = 1040 and is_obsolete = 0.

src/main/resources/liquibase/crop_changelog/19_4_0.xml
```diff
@@ -0,0 +1,37 @@
...
+		</addColumn>
+    </changeSet>
+
+	<changeSet author="cuenyad" id="v19.4.0-2">
+		<preConditions onFail="MARK_RAN">
+			<columnExists tableName="cvterm" columnName="is_system"/>
+		</preConditions>
+		<comment>Mark system variables</comment>
+        <sql>
+			update cvterm cv INNER JOIN cvterm cv1 ON cv.cvterm_id = cv1.cvterm_id  set cv.is_system = 1
```
clarysabel, line 27: why an inner join is needed? in both cases it does not seem to be necessary! @cuenyad

PR #1190, opened 2021-12-14
IBP-5289: merge germplasm - transfer associated files to target germp…

lik-83:
…lasm


nahuel-soldevilla (approved):
Looks good to me. 
I'll just mention here what you first mentioned and we discussed later, that there's no need to change the path in the bucket if we think of it as permanent key identifier for the object (even if it has the guuid in the name), which also simplifies the implementation a lot because we don't need to move things around and rename paths in the file metadata.

nahuel-soldevilla (approved):

PR #1198, opened 2021-12-30
IBP-4838 IBP-5377 manage program settings beta

cuenyad:
Please, review

Included:
https://github.com/IntegratedBreedingPlatform/BMSAPI/pull/642
https://github.com/IntegratedBreedingPlatform/Commons/pull/334
https://github.com/IntegratedBreedingPlatform/Fieldbook/pull/1050
https://github.com/IntegratedBreedingPlatform/Workbench/pull/621


src/main/java/org/generationcp/middleware/api/breedingmethod/BreedingMethodDTO.java
```diff
@@ -1,15 +1,23 @@
...
+import java.util.Date;
+import java.util.List;
+
 @AutoProperty
 public class BreedingMethodDTO extends BreedingMethodNewRequest {
 
 	private Integer mid;
 	private Boolean isBulkingMethod;
 	private Boolean isFavorite;
+	private Date date;
```
clarysabel, line 17: @cuenyad should we add JsonFormat to this? 
I think we should rename it to something more representative like creationDate

src/main/resources/liquibase/workbench_changelog/20_1_0.xml
```diff
@@ -0,0 +1,54 @@
...
+		<sql dbms="mysql" splitStatements="true">
+			INSERT INTO workbench_tool (`name`, `group_name`, `title`, `version`, `tool_type`, `path`, `parameter`, `user_tool`)
+			VALUES ('program_settings_manager', 'program_settings_manager', 'Manage Program Settings Beta', '1.0', 'WEB', '/ibpworkbench/controller/jhipster#program-settings-manager', '', 0);
+
+			INSERT INTO workbench_sidebar_category_link (`tool_name`, `sidebar_category_id`, `sidebar_link_name`, `sidebar_link_title`, `rank`)
+			VALUES ('program_settings_manager', 7, 'program_settings_manager', 'Manage Program Settings Beta', 2);
+		</sql>
+	</changeSet>
+
+	<changeSet author="nahuel" id="v20.1.0-2">
```
clarysabel, line 25: @nahuel-soldevilla this is only for testing right? We will need to add another changeset to delete this permission when we disable the old module

nahuel-soldevilla: posting here what we discussed:

only for testing, yes.
About **selectable**, yes it's possible but we can also disregard any intermediate state when we do the final migration, and copy the exact same state as the old permission when we disable it (the roles that have the old permission are automatically given the new one)
This also give us more opportunities to test it exactly as it's going to be in the end, as a selectable permission.
Any intermediate association with the the beta permission will be disregarded later.

The **Api**/**PreAuthorize** can be tested by assigning the final permission name (MANAGE_PROGRAM_SETTINGS instead of MANAGE_PROGRAM_SETTINGS_BETA) and giving both permission through the side admin, one for accessing the module in the side menu, and the other for giving access to the api calls. This way in the final migration we don't need to rename permission in the the preauthorize.

clarysabel: I think we can simplify the changeset if we are going to disable the old module in 20.1 
If testing passes and before merging we can discuss with Mariano and decide how to proceed
@nahuel-soldevilla

src/main/java/org/generationcp/middleware/api/location/LocationService.java
```diff
@@ -38,4 +38,10 @@
 	LocationDTO createLocation(LocationRequestDto locationRequestDto);
 
 	void updateLocation(Integer locationId, LocationRequestDto locationRequestDto);
+
+	boolean existsLocationAsCountry(Integer locationId);
```
clarysabel, line 42: @cuenyad rename to isDefaultCountryLocation

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
```diff
@@ -61,6 +64,27 @@ public long countFilteredLocations(final LocationSearchRequest locationSearchReq
 	@Override
 	public void deleteLocation(final Integer locationId) {
 		final Location location = this.daoFactory.getLocationDAO().getById(locationId);
+		List<Locdes> Blocks = null;
```
clarysabel, line 67: @cuenyad rename to blocks

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
```diff
@@ -61,6 +64,27 @@ public long countFilteredLocations(final LocationSearchRequest locationSearchReq
...
+		if (location.getLtype() == 415) { // Field
```
clarysabel, line 68: @cuenyad create (or find an existing place) an enum (or use constants) for 415 and 416, and also use it in BMSAPI because the values are also used there

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
```diff
@@ -61,6 +64,27 @@ public long countFilteredLocations(final LocationSearchRequest locationSearchReq
...
+			//Get the Blocks
+			Blocks = daoFactory.getLocDesDao().getLocdes(null, locationId.toString());
+			Blocks.forEach((locdes) -> {
+					this.daoFactory.getLocDesDao().makeTransient(locdes);
```
clarysabel, line 72: @cuenyad sounds like we dont need to go n times to the database, we could have only 2 DELETE queries to perform these actions

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
```diff
@@ -61,6 +64,27 @@ public long countFilteredLocations(final LocationSearchRequest locationSearchReq
...
+					final Location blockLocation = this.daoFactory.getLocationDAO().getById(locdes.getLocationId());
+					this.daoFactory.getLocationDAO().makeTransient(blockLocation);
+				}
+			);
+
+			final List<Locdes> fields = daoFactory.getLocDesDao().getLocdes(locationId, null);
+			fields.forEach(locdes -> this.daoFactory.getLocDesDao().makeTransient(locdes));
```
clarysabel, line 79: @cuenyad same here, avoid calling a makeTransient with a forEach, if we can have only one DELETE action

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
```diff
@@ -61,6 +64,27 @@ public long countFilteredLocations(final LocationSearchRequest locationSearchReq
...
+		}
+		if (location.getLtype() == 416) { // Block
+			Blocks = daoFactory.getLocDesDao().getLocdes(locationId, null);
+			Blocks.forEach(locdes -> {
```
clarysabel, line 83: @cuenyad not clear to me why in this case we save locdes?

src/main/java/org/generationcp/middleware/dao/LocationDAO.java
```diff
@@ -589,4 +608,33 @@ public Location getDefaultLocationByType(final Integer type) {
 		return Optional.empty();
 	}
 
+	public boolean blockIdIsUsedInFieldMap(final List<Integer> blockIds) {
+		try {
+
+			final StringBuilder queryString = new StringBuilder().append("SELECT ");
```
clarysabel, line 614: @cuenyad 1) use a constant 2) dont we have constants fro 1155 and 8583?

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
cuenyad: @clarysabel per field you can have more than one block.

cuenyad: done.

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
cuenyad: done.

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
cuenyad: done.

src/main/java/org/generationcp/middleware/api/location/LocationService.java
cuenyad: done.

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
cuenyad: done.

src/main/java/org/generationcp/middleware/dao/LocationDAO.java
cuenyad: done.

src/main/java/org/generationcp/middleware/api/location/LocationServiceImpl.java
cuenyad: done.

PR #1208, opened 2022-01-13
IBP-5296: add 'entry_code' column as default

lik-83:


src/main/resources/liquibase/crop_master.xml
```diff
@@ -83,6 +83,7 @@
 	<include file="crop_changelog/19_3_0.xml" relativeToChangelogFile="true"/>
 	<include file="crop_changelog/19_4_0.xml" relativeToChangelogFile="true"/>
 	<include file="crop_changelog/19_4_3.xml" relativeToChangelogFile="true"/>
+	<include file="crop_changelog/19_5_0.xml" relativeToChangelogFile="true"/>
```
clarysabel, line 86: @lik-83 a small detail, if this passes branch testing we will merge it as part of 19.4.3
so we will have to move it to that xml file.

lik-83: Done

PR #1213, opened 2022-01-19
IBP-5326: migrate 'listdata.source' values as a new entry detail

lik-83:


src/main/resources/liquibase/crop_changelog/19_5_0.xml
```diff
@@ -20,4 +20,39 @@
 		</sql>
 	</changeSet>
 
+	<changeSet author="lkovacic" id="v19.5.0-2">
+		<preConditions onFail="MARK_RAN">
+			<and>
+				<sqlCheck expectedResult="0">select count(*) from cvterm where name = 'ENTRY_SOURCE';</sqlCheck>
```
clarysabel, line 26: We should check by cv_id=1040  and is_obsolete = 0 given that's the unique key of the table 
@lik-83

src/main/resources/liquibase/crop_changelog/19_5_0.xml
```diff
@@ -20,4 +20,39 @@
...
+			INSERT INTO cvterm_relationship(type_id, subject_id, object_id)
+			VALUES
+			(1200, @cvterm_id, 2208),
+			(1210, @cvterm_id, 4030),
+			(1220, @cvterm_id, 6020);
+
+			INSERT INTO cvtermprop(cvterm_id, type_id, value, rank)
+			VALUES
+			(@cvterm_id, 1800,'Entry Detail', 0),
+			(@cvterm_id, 8048, (SELECT DATE_FORMAT(CURRENT_TIMESTAMP(), '%Y-%m-%dT%T.000Z')), 0);
```
clarysabel, line 46: @lik-83 please check creating a new variable via ontology manager and confirm if we are always setting rank = 0 o if we should use 1 in the second insert

lik-83: I already checked that. Both ranks are equals to 0 when an entry detail variable is created.

src/main/resources/liquibase/crop_changelog/19_5_0.xml
lik-83: done

src/main/resources/liquibase/crop_changelog/19_5_0.xml
```diff
@@ -20,4 +20,39 @@
...
+		</preConditions>
+
+		<comment>Add "ENTRY_SOURCE" variable and migrate 'listdata.source' values as list entry details</comment>
+        <sql dbms="mysql" splitStatements="true">
+			INSERT INTO cvterm (cv_id, name, definition, dbxref_id, is_obsolete, is_relationshiptype)
+			VALUES (1040, 'ENTRY_SOURCE', 'Entry Source', null, 0, 0);
+
+			SET @cvterm_id  = (SELECT LAST_INSERT_ID());
+
+			INSERT INTO cvterm_relationship(type_id, subject_id, object_id)
```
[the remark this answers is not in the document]
clarysabel: Hi @mchabito I had a conversation with @mcrimi about this specific case. 
He said the probabilities are low so we will proceed with branch testing and merging. DM's will have to check the data if required.

nahuel-soldevilla (approved):

PR #1216, opened 2022-01-27
IBP-5424 set historical list folders to have lhierarchy=null

mchabito:
For review and merge please, @clarysabel . Thanks!

### DIFF
--- a/src/main/resources/liquibase/crop_changelog/19_5_1.xml
+++ b/src/main/resources/liquibase/crop_changelog/19_5_1.xml
@@ -4,26 +4,39 @@
 				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
 		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
 	
-		<changeSet author="rowena" id="v19.5.1-1">
+	<changeSet author="rowena" id="v19.5.1-1">
         <preConditions onFail="MARK_RAN">
             <sqlCheck expectedResult="1">
                 SELECT COUNT(*) > 0 FROM germplsm g INNER JOIN location l1  INNER JOIN location l2  WHERE  g.glocn = l1.locid  AND l1.locid > l2.locid AND l1.lname = l2.lname AND l1.ltype= l2.ltype AND l1.cntryid = l2.cntryid AND l1.locid=l1.snl1id
               </sqlCheck>
         </preConditions>
-        <comment> check if there are germplasms using duplicate locatio_id;  Replace the duplicates with the GID of the minimum location ID of the duplicate</comment>
+        <comment> check if there are germplasms using duplicate location_id. Replace the duplicates with the locid of the minimum location ID of the duplicate</comment>
 		<sql dbms="mysql" splitStatements="true">
 	      	UPDATE germplsm g INNER JOIN location l1  INNER JOIN location l2   SET glocn = l2.locid   WHERE  g.glocn = l1.locid  AND l1.locid < l2.locid AND l1.lname = l2.lname  AND l1.ltype= l2.ltype AND l1.cntryid = l2.cntryid AND l1.locid=l1.snl1id ;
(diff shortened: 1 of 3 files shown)


src/main/resources/liquibase/crop_changelog/19_5_1.xml
```diff
@@ -0,0 +1,57 @@
+﻿<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+				   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+	
+	<changeSet author="rowena" id="v19.5.1-1">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="1">
+                SELECT COUNT(*) > 0
```
darla-leafnode, line 10: Hi @mchabito, @rvtulod - my concern is it's only checking and updating usage of the "dupe" location in germplsm table. However, there are several other places where a location could be referred:
- atributs.alocn
- instln.ulocn (not sure)
 - names.nlocn
- study variables (mainly nd_geolocation but also possible in projectprop,phenotype when a variable has location data type..)

I think we should be checking/updating usage of those too before we delete any location to avoid orphan records.

src/main/resources/liquibase/crop_changelog/19_5_1.xml
```diff
@@ -0,0 +1,57 @@
...
+						 INNER JOIN location l1
+						 INNER JOIN location l2
+				WHERE g.glocn = l1.locid AND l1.locid > l2.locid AND l1.lname = l2.lname AND l1.ltype = l2.ltype AND l1.cntryid = l2.cntryid AND l1.locid = l1.snl1id
+              </sqlCheck>
+        </preConditions>
+        <comment> check if there are germplasm using duplicate location_id. Replace the duplicates with the locid of the minimum location ID of the duplicate</comment>
+		<sql dbms="mysql" splitStatements="true">
+	      	UPDATE germplsm g INNER JOIN location l1 INNER JOIN location l2
+			SET glocn = l2.locid
+			WHERE g.glocn = l1.locid AND l1.locid &lt; l2.locid AND l1.lname = l2.lname AND l1.ltype= l2.ltype AND l1.cntryid = l2.cntryid AND l1.locid=l1.snl1id;
```
darla-leafnode, line 21: I notice the update statement use `&lt;` instead of ">" - which is used in the precondition

src/main/resources/liquibase/crop_changelog/19_5_1.xml
```diff
@@ -0,0 +1,57 @@
...
+		</sql>
+	</changeSet>
+	
+	<changeSet author="rowena" id="v19.5.1-2">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="1">
+                SELECT COUNT(*) > 0
+				FROM location l1
+						 INNER JOIN location l2
+				WHERE l1.locid &lt; l2.locid AND l1.lname = l2.lname AND l1.ltype = l2.ltype AND l1.cntryid = l2.cntryid AND l1.locid = l1.snl1id;
```
darla-leafnode, line 31: I notice the preconditions and delete statements use `&lt;`

darla-leafnode (changes requested):

darla-leafnode:
Hi @mchabito and @rvtulod - please see my comments/concerns. On changesets 1 and 2, I think we should not rush merging this change and carefully consider and test where else these locations could have been used.

On changeset 3, I think it is okay. Is this a must for v20 in your end? You can create a separate branch with just this change (or remove changesets 1 and 2 from this branch).

cc; @clarysabel , @christineguiang

PR #1240, opened 2022-03-07
IBP-5481 Import Means Data

abatac:


src/main/java/org/generationcp/middleware/service/impl/analysis/MeansImportRequest.java
```diff
@@ -0,0 +1,59 @@
...
+		return this.data;
+	}
+
+	public void setData(final List<MeansData> data) {
+		this.data = data;
+	}
+
+	public static class MeansData {
+
+		private Integer environmentId;
```
darla-leafnode, line 29: I think using `environmentNumber` (ie. trial instance 1, 2, 3..) will make more sense here than the environmentId (ie. nd_geolocation_id) because I think most if not all apps will not have straightforward access to the environment IDs - which are internal. 

In BV, the trial environment factor is usually the trial instance number of location name hence the result usually uses either of the two as trial environment identifier. I suspect the same for StABrAPP (please check)

darla-leafnode (changes requested):

src/main/java/org/generationcp/middleware/manager/api/StudyDataManager.java
```diff
@@ -35,6 +35,7 @@
 import org.generationcp.middleware.pojos.Method;
 import org.generationcp.middleware.pojos.dms.DmsProject;
 import org.generationcp.middleware.pojos.dms.PhenotypeOutlier;
+import org.generationcp.middleware.pojos.dms.StockModel;
```
darla-leafnode, line 38: Remove unused import

src/main/java/org/generationcp/middleware/service/api/analysis/SiteAnalysisService.java
```diff
@@ -0,0 +1,8 @@
+package org.generationcp.middleware.service.api.analysis;
+
+import org.generationcp.middleware.service.impl.analysis.MeansImportRequest;
+
+public interface SiteAnalysisService {
+
+	Integer createMeansDataset(final Integer studyId, MeansImportRequest meansRequestDto);
```
darla-leafnode, line 7: Remove final

darla-leafnode (approved):

PR #1250, opened 2022-04-18
IBP-5555: adjust germplasm name max length

cheryleigh:


src/main/resources/liquibase/crop_changelog/20_4_0.xml
```diff
@@ -0,0 +1,115 @@
...
+				SELECT COUNT(1)
+				FROM INFORMATION_SCHEMA.COLUMNS c
+				WHERE TABLE_NAME = 'listdata'
+				AND COLUMN_NAME = 'grpname'
+				AND CHARACTER_MAXIMUM_LENGTH = 255
+				AND TABLE_SCHEMA = DATABASE();
+			</sqlCheck>
+		</preConditions>
+		<comment>
+			Extend length of stock name column
```
abatac, line 109: Change to "Extend length of stock grpname column"

PR #1277, opened 2022-05-24
IBP-5633 clean up

mderamos:


src/main/resources/liquibase/workbench_changelog/21_1_0.xml
```diff
@@ -0,0 +1,37 @@
...
+		<preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="1">
+				SELECT count(*) FROM workbench_sidebar_category_link WHERE tool_name = 'manage_program';
+			</sqlCheck>
+		</preConditions>
+		<comment>
+			Remove old 'Manage Program Settings' link
+		</comment>
+		<sql dbms="mysql" splitStatements="true">
+			DELETE FROM workbench_sidebar_category_link
```
clarysabel, line 17: Please consider the scenario  where a DM enabled the old module using the id in permissions table. 
In that case I would replace the old id by the new one
Finally delete in workbench_sidebar_category_link 
@mderamos

PR #1278, opened 2022-05-26
IBP-5541 program location default

mderamos:


src/main/java/org/generationcp/middleware/dao/ProgramLocationDefaultDAO.java
```diff
@@ -0,0 +1,40 @@
...
+import org.hibernate.Session;
+import org.hibernate.criterion.Restrictions;
+
+public class ProgramLocationDefaultDAO extends GenericDAO<ProgramLocationDefault, Integer> {
+
+	public ProgramLocationDefaultDAO(final Session session) {
+		super(session);
+	}
+
+	public ProgramLocationDefault getByprogramUUID(final String programUUID) throws MiddlewareQueryException {
```
clarysabel, line 17: @mderamos use camelCase getByprogramUUID by getByProgramUUID

src/main/resources/liquibase/crop_changelog/21_1_0.xml
```diff
@@ -22,4 +22,64 @@
...
+					</sqlCheck>
+				</not>
+			</and>
+		</preConditions>
+		<comment>
+			Populate program_location_default for existing programs, set 0 as default
+		</comment>
+		<sql dbms="mysql" splitStatements="true">
+			INSERT INTO program_location_default(program_uuid, location_id)
+			SELECT project.project_uuid, 0 FROM workbench.workbench_project project
```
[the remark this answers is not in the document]
mderamos: Hi @clarysabel , I just noticed that it's not a deletable location since its ltype = 405. There's a validation where the system prevents the users from deleting locations with ltype=405.

But as a safety precaution, I also added a liquibase script that checks if the location with labbr UKN is existing. If not, reinsert it in the table and use the reinserted value as the default for existing programs.

src/main/resources/liquibase/crop_changelog/21_1_0.xml
```diff
@@ -22,4 +22,80 @@
...
 	</changeSet>
 
+	<changeSet author="mderamos" id="v21.1.0-2">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<tableExists tableName="program_location_default"/>
+			</not>
+		</preConditions>
+		<comment>
+			Create program_location_default table
```
clarysabel, line 32: Hi @mderamos 
I have a concern, is there a reason to create a new table to store default locations that are unique per program? Should not be more accurate to create columns y workbench_project? 
I think this simplify the code

PR #1283, opened 2022-06-03
IBP-5714 storage location default

mderamos:


src/main/resources/liquibase/crop_changelog/21_2_0.xml
```diff
@@ -0,0 +1,80 @@
...
+			<column name="storage_location_id" type="INT(11)" afterColumn="breeding_location_id">
+				<constraints nullable="false" foreignKeyName="fk_program_storage_location_default" references="location(locid)"/>
+			</column>
+		</addColumn>
+	</changeSet>
+
+	<changeSet author="mderamos" id="v22.1.0-3">
+		<preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="0">
+				SELECT COUNT(*) FROM location WHERE locid = 6000 AND ltype = 1500;
```
clarysabel, line 37: Should we check that DSS is not used as an abbreviation? 
@mderamos

PR #1308, opened 2022-07-29
IBP-5770: implement import entry details in germplasm and checks

cheryleigh:


src/main/java/org/generationcp/middleware/api/study/StudyEntryObservationServiceImpl.java
```diff
@@ -32,12 +33,18 @@ public Integer createObservation(final StockPropertyData stockPropertyData) {
...
-		final StockProperty stockProperty =
-			this.daoFactory.getStockPropertyDao().getByStockIdAndTypeId(stockPropertyData.getStockId(), stockPropertyData.getVariableId())
-				.orElseThrow(() -> new MiddlewareException(
-					"Stock property: " + stockPropertyData.getStockId() + " with type: " + stockPropertyData.getVariableId()
-						+ " not found."));
+	public Integer updateObservation(final StockPropertyData stockPropertyData, final boolean allowCreate) {
+		final Optional<StockProperty> stockPropertyOptional = this.daoFactory.getStockPropertyDao().getByStockIdAndTypeId(
+			stockPropertyData.getStockId(), stockPropertyData.getVariableId());
+
+		if (!stockPropertyOptional.isPresent() && allowCreate) {
```
abatac, line 40: Move the logic that checks for allowCreate and creation of observation outside of this method, so that it's clear that this is just for updating of observation.

abatac (changes requested):

src/main/java/org/generationcp/middleware/api/study/StudyEntryObservationServiceImpl.java
cheryleigh: created new method: createOrUpdateObservation

abatac (approved):

PR #1324, opened 2022-08-10
IBP-5730: implement fieldmap deletion

cheryleigh:


src/main/java/org/generationcp/middleware/dao/dms/GeolocationPropertyDao.java
```diff
@@ -102,18 +102,22 @@ public List<Integer> deleteBlockPropertiesByGeolocationId(final List<Integer> ge
 
 			// block IDs to return to be used for locdes deletion
 			// only return if no other geolocation uses the block
-			SQLQuery sqlQuery1 = this.getSession().createSQLQuery(GET_BLOCK_IDS_TO_DELETE);
+			SQLQuery sqlQuery1 = this.getSession().createSQLQuery(GET_BLOCK_IDS_TO_DELETE)
```
clarysabel, line 105: This is not very easy to understand, it was not in fact. 
This is only supposed to delete some rows but it is returning some blocks that are not even used internally in the delete script. 
why dont we split this in 2 functions? it is not easy to undertand what the map repesents 
I would find the affected blocks before calling the deletion and if it is out of scope, then much better to return ids and then out of this function do whatever you need to get the affected objects. 
@cheryleigh

src/main/java/org/generationcp/middleware/dao/dms/GeolocationDao.java
```diff
@@ -485,6 +486,14 @@ public Integer getNextInstanceNumber(final Integer datasetId) {
 		return 1;
 	}
 
+	public List<String> getInstanceNumbersFromIds (final List<Integer> instanceIds) {
```
clarysabel, line 489: @cheryleigh Thinking about reusability, I think it will be more useful to implement getByIds, returning a list of Geolocation and then build the list out of the DAO

PR #1327, opened 2022-08-12
IBP-5644 add sname type

mderamos:


src/main/java/org/generationcp/middleware/api/breedingmethod/BreedingMethodServiceImpl.java
```diff
@@ -111,10 +135,25 @@ public BreedingMethodDTO edit(final Integer breedingMethodDbId, final BreedingMe
...
+		} else {
+			// Set to null for scenario where the user removed the snametype value
+			method.setSnametype(null);
+		}
+
 		final BreedingMethodMapper mapper = new BreedingMethodMapper();
 		mapper.mapForUpdate(breedingMethod, method);
 		this.daoFactory.getMethodDAO().update(method);
-		return new BreedingMethodDTO(method);
+		final BreedingMethodDTO breedingMethodDTO = new BreedingMethodDTO(method);
```
clarysabel, line 151: @mderamos I would also add snameTypeId to the response in case we want to redirect to the name type edition for example. 
I would suggest to have id in the request and then name in the DTO

src/main/java/org/generationcp/middleware/api/breedingmethod/BreedingMethodServiceImpl.java
```diff
@@ -63,6 +71,9 @@ public Optional<BreedingMethodDTO> getBreedingMethod(final Integer breedingMetho
 		if (!Objects.isNull(methodEntity)) {
 			final BreedingMethodDTO breedingMethodDTO = new BreedingMethodDTO(methodEntity);
 			breedingMethodDTO.setFavorite(programOptional.isPresent() ? programFavorite.isPresent() : null);
+			if (methodEntity.getSnametype() != null) {
+				breedingMethodDTO.setSnameTypeCode(this.daoFactory.getUserDefinedFieldDAO().getById(methodEntity.getSnametype()).getFcode());
```
clarysabel, line 75: @mderamos setting ids could help in the future

PR #1344, opened 2022-08-31
Fix precondition to prevent the script to run in UAT and BrAPI server…

clarysabel:
…s where the data will be fixed manually

IBP-5864

### DIFF
--- a/src/main/resources/liquibase/crop_changelog/21_3_0.xml
+++ b/src/main/resources/liquibase/crop_changelog/21_3_0.xml
@@ -151,16 +151,21 @@
 		</sql>
 	</changeSet>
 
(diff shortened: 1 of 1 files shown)


nahuel-soldevilla (approved):

clarysabel:
Context: 

We have run a wrong changeset in all environments (centos, UAT and BrAPI servers)
Unfortunately for UAT and BrAPI servers we dont have a recent database backup for all crops installed there. 
For that reason, we have to fix the data manually.

To prevent the changeset to run twice in the environments where the data has been manually fixed, I added an <and> to the precondition, so it will only run when the Entry type is still a Germplasm Descriptor. 

This <and> condition will not affect any other server installation where this variable is supposed to be Germplasm Descriptor.

PR #1350, opened 2022-09-09
IBP-4433 Removed Germplasm.methodId field

mderamos:


src/main/java/org/generationcp/middleware/pojos/Germplasm.java
```diff
@@ -477,9 +446,8 @@ public Germplasm(final Integer gid, final Integer methodId, final Integer gnpgs,
 
 	public Germplasm(final Integer gid, final Integer methodId, final Integer gnpgs, final Integer gpid1, final Integer gpid2,
 		final Integer lgid, final Integer locationId, final Integer gdate, final Name preferredName) {
-
 		// gref =0, grplce = 0, mgid = 0
-		this(gid, methodId, gnpgs, gpid1, gpid2, lgid, locationId, gdate, 0, 0, 0, preferredName, null, null);
+		this(gid, gnpgs, gpid1, gpid2, lgid, locationId, gdate, 0, 0, 0, preferredName, null, new Method(methodId));
```
abatac, line 450: if methodId is null, just pass null, not new method with null methodId

abatac (changes requested):

PR #1354, opened 2022-09-14
IBP-5872: add breeding method modal on click of field in germplasm and checks

cheryleigh:


src/main/java/org/generationcp/middleware/dao/study/StudyEntrySearchDAO.java
```diff
@@ -280,6 +291,31 @@ private void addGuidScalar(final List<Scalar> scalars, final List<String> select
 		}
 	}
 
+	private void addBreedingMethodIdScalar(final List<Scalar> scalars, final List<String> selectClause,
+		final List<MeasurementVariable> entryDescriptors) {
+		if (!CollectionUtils.isEmpty(entryDescriptors)) {
+			// automatically add breeding method ID if method abbr is present (required for breeding method modal link)
```
clarysabel, line 297: @cheryleigh can the modal be implemented using abbreviation if this is considered unique? so we keep this code simple and easy to maintain.

src/main/java/org/generationcp/middleware/service/api/study/StudyEntryDto.java
```diff
@@ -34,6 +34,8 @@ public class StudyEntryDto implements Serializable {
 
 	private String guid;
 
+	private String breedingMethodId;
```
clarysabel, line 37: @cheryleigh if we use abbreviation for implementing the popup we dont need to add this attribute, right?

src/main/java/org/generationcp/middleware/dao/study/StudyEntrySearchDAO.java
cheryleigh: hi @clarysabel , yup, that is indeed a better approach. I updated the modal to handle retrieval from abbreviation, thank you!

PR #1364, opened 2022-10-03
IBP-5823-DeleteValueBatchAction

mderamos:


src/main/java/org/generationcp/middleware/service/impl/dataset/DatasetServiceImpl.java
```diff
@@ -1121,6 +1108,54 @@ public void setValueToVariable(final Integer datasetId, final ObservationUnitsPa
...
+				if (observationUnitData != null) {
+					phenotype = this.daoFactory.getPhenotypeDAO().getById(observationUnitData.getObservationId());
+				}
+
+				if (phenotype != null) {
+					if (draftMode) {
+						if(phenotype.getcValueId() == null && StringUtils.isEmpty(phenotype.getValue())) {
+							phenotypeIdsToDelete.add(phenotype.getPhenotypeId());
+						} else {
+							this.updatePhenotype(phenotype,
```
clarysabel, line 1135: @mderamos seems to me that it is possible to run only one update instead of calling updates n-times

src/main/java/org/generationcp/middleware/service/impl/dataset/DatasetServiceImpl.java
```diff
@@ -1121,6 +1108,54 @@ public void setValueToVariable(final Integer datasetId, final ObservationUnitsPa
...
+								phenotype.getcValueId(), phenotype.getValue(), null, null, true);
+						}
+					} else {
+						if(phenotype.getDraftCValueId() == null && StringUtils.isEmpty(phenotype.getDraftValue())) {
+							phenotypeIdsToDelete.add(phenotype.getPhenotypeId());
+						} else {
+							this.updatePhenotype(phenotype,
```
clarysabel, line 1142: @mderamos same here, it should be possible to execute one update

src/main/java/org/generationcp/middleware/service/api/dataset/ObservationUnitsSearchDTO.java
```diff
@@ -52,6 +52,7 @@ public class Filter {
 		private Set<Integer> filteredNdExperimentIds;
 		private Map<String, String> variableTypeMap;
 		private Integer variableId;
+		private Boolean variableHasValue;
```
clarysabel, line 55: @mderamos what exactly means variableHasValue? has value? has cValue_id? hasPhenotype?

mderamos: Hi @clarysabel , 
**what exactly means variableHasValue? has value? has cValue_id? hasPhenotype?**
it's either has value or has draftValue depending on the draft mode. It's used for the prompt message showing how many observations will be affected with the delete variable values batch action. Only those with values(or draft values depending on the draft mode) should be filtered. It's also used for the actual deletion method so that the query will only return the necessary observations. 

**what happens if draftMode is TRUE and variableHasValue is TRUE?**
if the draftmode is true, it already filters the observations with non-null draft value so it will just ignore the _variableHasValue _ filter.

**It may be possible to set this value in the REST service and not in the frontend**
Done addressing this.

PR #1367, opened 2022-10-05
IBP-2676: implement handling of obsolete variables

cheryleigh:


src/main/resources/liquibase/crop_changelog/22_2_0.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+				   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+	<changeSet author="cheryleigh" id="v22.2.0-1">
+		<preConditions onFail="MARK_RAN">
+			<indexExists tableName="cvterm" indexName="cvterm_idx1"/>
```
clarysabel, line 9: @cheryleigh createIndex will fail if we have 2 cvterms with same name anc cv_id and different is_obsolete. Please add a precondition to check that condition

clarysabel (changes requested):
see PR comment

PR #1370, opened 2022-10-13
IBP-5777 IBP-5942 IBP-5981 Import Pedigree Brapi Sync

abatac:


src/main/java/org/generationcp/middleware/dao/GermplasmDAO.java
```diff
@@ -612,7 +612,6 @@ public List<PedigreeNodeDTO> searchPedigreeNodes(final PedigreeNodeSearchRequest
 		final GermplasmSearchRequest germplasmSearchRequest = this.convertToGermplasmSearchRequest(pedigreeNodeSearchRequest);
 		final SQLQuery filterQuery = this.getSession().createSQLQuery(this.buildFilterGermplasmQuery(germplasmSearchRequest));
 		this.addGermplasmSearchParameters(new SqlQueryParamBuilder(filterQuery), germplasmSearchRequest);
-		addPaginationToSQLQuery(filterQuery, pageable);
```
nahuel-soldevilla, line 615: What's the rationale for this? (removing pagination)
I guess it's tricky to combine in this case pagination with pedigree + pedigree levels, and it doesn't seem to be well defined in the brapi specs how it should work in this case, at least nothing is mentioned here: https://plant-breeding-api.readthedocs.io/en/latest/docs/best_practices/Pedigree_Tree.html
One possible approach is the one of bmsapi germplasm search, where the page applies only to the root nodes, and the parents are "extra" items that don't add up to the page size.

Another option would be to send a very large page size from brapi-sync tool, if we need to get all.

But I guess this would work for now, maybe leave a couple of notes or TODOs to indicate why the page is being ignored, also probably at the api level with api docs, to make users aware.

abatac: Hi @nahuel-soldevilla  I removed the pagination here so that if client searches for more than 1000 germplasm it will return all the germplasm in the result. It's necessary to get all the root germplasm here so that we can process them for pedigree extraction.

I tried to specify the pageSize from the brapi-sync tool but there's a 5000 validation limit.

PR #1380, opened 2022-11-04
IBP-5874 Implement Two Factor Authentication

abatac:


src/main/resources/liquibase/workbench_changelog/22_3_0.xml
```diff
@@ -0,0 +1,55 @@
...
+				<tableExists tableName="one_time_password"></tableExists>
+			</not>
+		</preConditions>
+		<comment>
+			Create table 'one_time_password'
+		</comment>
+		<sql dbms="mysql" splitStatements="true">
+			CREATE TABLE `workbench`.`one_time_password` (
+				`id`       INT NOT NULL AUTO_INCREMENT,
+				`otp_code` INT NULL,
```
clarysabel, line 18: @abatac can otp_code and expires be null?

src/main/resources/liquibase/workbench_changelog/22_3_0.xml
```diff
@@ -0,0 +1,55 @@
...
+				<tableExists tableName="user_device_metadata"></tableExists>
+			</not>
+		</preConditions>
+		<comment>
+			Create table 'user_device_meta_data'
+		</comment>
+		<sql dbms="mysql" splitStatements="true">
+		  CREATE TABLE `workbench`.`user_device_meta_data` (
+			  `id`             INT NOT NULL AUTO_INCREMENT,
+			  `userid`         INT NULL,
```
clarysabel, line 36: @abatac seems userid should not be null and have a FK to users

src/main/resources/liquibase/workbench_changelog/22_3_0.xml
```diff
@@ -0,0 +1,55 @@
...
+	</changeSet>
+	<changeSet author="aldrin" id="v22.3.0-3">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<columnExists tableName="users" columnName="multi_factor_authentication"/>
+			</not>
+		</preConditions>
+		<sql dbms="mysql" splitStatements="true">
+			ALTER TABLE `workbench`.`users`
+				ADD COLUMN `multi_factor_authentication` TINYINT NULL DEFAULT 0 AFTER `cdate`;
```
clarysabel, line 52: @abatac when reading first time, I wasnt hundred percent clear, I think multi_factor_enabled represents better the stored value

nahuel-soldevilla: or a little bit longer.. `multi_factor_authentication_enabled`
it could also be `multi_factor_enabled` with a column `message` for more info

pom.xml
```diff
@@ -364,6 +364,21 @@
 			<version>2.0</version>
 		</dependency>
 
+		<!-- Dependency for passing browser/client's User-Agent -->
```
nahuel-soldevilla, line 367: parsing

src/main/java/org/generationcp/middleware/util/UserDeviceMetaDataUtil.java
```diff
@@ -0,0 +1,43 @@
...
+		if (Objects.nonNull(userAgent)) {
+			return userAgent.getBrowser()
+				+ " " + userAgent.getBrowserVersion().getMajorVersion() + "."
+				+ userAgent.getBrowserVersion().getMinorVersion() + " - "
+				+ userAgent.getOperatingSystem().getDeviceType() + " - "
+				+ userAgent.getOperatingSystem().getName();
+		} else {
+			return UNKNOWN;
+		}
+	}
```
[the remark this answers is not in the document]
abatac: This is a good idea. I think we can just store the whole User-Agent string extracted from the request header. And if we want to display it, we can just parse it.

src/main/resources/liquibase/workbench_changelog/22_3_0.xml
abatac: Those columns will never be null. I updated them to NOT NULL

src/main/resources/liquibase/workbench_changelog/22_3_0.xml
abatac: Done.

src/main/resources/liquibase/workbench_changelog/22_3_0.xml
abatac: I changed the column name to multi_factor_authentication_enabled

pom.xml
abatac: Done

src/main/java/org/generationcp/middleware/util/UserDeviceMetaDataUtil.java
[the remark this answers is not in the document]
abatac: Done

PR #1383, opened 2022-11-08
IBP-4828-Delete-UserDetailsDto

clarysabel:
IBP-4828


nahuel-soldevilla (approved):
Looks good. As you mentioned, we can address the "status as a string" issue later with the site admin refactor

PR #1389, opened 2022-11-15
IBP-6074-Delete-addUser

clarysabel:
IBP-6074

### DIFF
--- a/src/main/java/org/generationcp/middleware/service/impl/user/UserServiceImpl.java
+++ b/src/main/java/org/generationcp/middleware/service/impl/user/UserServiceImpl.java
@@ -35,6 +35,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 @Service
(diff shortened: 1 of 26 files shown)


src/main/java/org/generationcp/middleware/service/impl/user/UserServiceImpl.java
```diff
@@ -74,8 +72,12 @@ public List<WorkbenchUser> getUserByName(final String name, final int start, fin
...
 
 	@Override
-	public WorkbenchUser getUserByFullname(final String fullname) {
-		return this.workbenchDaoFactory.getWorkbenchUserDAO().getUserByFullName(fullname);
+	public UserDto getUserByFullname(final String fullname) {
+		final WorkbenchUser user = this.workbenchDaoFactory.getWorkbenchUserDAO().getUserByFullName(fullname);
+		if (user != null) {
+			return new UserDto(user);
+		}
+		return null;
```
nahuel-soldevilla, line 80: maybe we can return Optional here

nahuel-soldevilla (approved):

PR #1398, opened 2022-11-22
Ibp 5941 implement post advance studies

lik-83:


lik-83:
This PR is no longer valid. A PR will be created for implementation of advancing study and samples.

PR #1399, opened 2022-11-22
IBP-6078-Fix-Site-Admin-Backend

clarysabel:
…id exposing hibernate entities in BMSAPI

IBP-6078

### DIFF
--- a/src/main/java/org/generationcp/middleware/service/api/permission/PermissionServiceImpl.java
+++ b/src/main/java/org/generationcp/middleware/service/api/permission/PermissionServiceImpl.java
@@ -6,8 +6,10 @@ import org.generationcp.middleware.hibernate.HibernateSessionProvider;
 import org.generationcp.middleware.manager.WorkbenchDaoFactory;
 import org.generationcp.middleware.pojos.workbench.Permission;
 import org.generationcp.middleware.pojos.workbench.RoleTypePermission;
+import org.generationcp.middleware.service.api.user.RoleTypeDto;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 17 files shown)


src/main/java/org/generationcp/middleware/service/api/permission/PermissionServiceImpl.java
```diff
@@ -56,19 +57,21 @@ private void getLinks(final Set<PermissionDto> permissionDtoList, final Permissi
...
+			final PermissionDto permissionDto = new PermissionDto(p);
+			final Map<RoleTypeDto, Boolean> roleTypeDtoSelectableMap = new HashMap<>();
+			p.getRoleTypePermissions().forEach(rtp -> {
+				roleTypeDtoSelectableMap.putIfAbsent(new RoleTypeDto(rtp.getRoleType()), rtp.getSelectable());
+			});
+			permissionDto.setRoleTypeSelectableMap(roleTypeDtoSelectableMap);
+			permissionDtoList.add(permissionDto);
+		}
+		);
+		return permissionDtoList;
```
nahuel-soldevilla, line 72: I feel like we can probably improve this, right now permissionDto.getRoleTypeSelectableMap is totally dependent on getPermissionsDtoByIds.
It is true that it is convenient because we are iterating once and building the map accordingly and for each permission, though from the method caller perspective this is not evident.

What about putting it as a separate service method instead of inside the dto?
```
Map<Integer, Map<RoleTypeDto, Boolean>> getRoleTypeSelectableByPermissionId()
```
or changing getPermissionsDtoByIds to make it more evident: `getPermissionsDtoByIdsWithRoleTypeSelectableMap`, a bit long but..
Or just a javadoc comment in each method stating the dependency..

nahuel-soldevilla (approved):

PR #1413, opened 2023-01-19
IBP-6118: add plot_no column to sub observations lists

cheryleigh:


src/main/java/org/generationcp/middleware/domain/sample/SampleDetailsDTO.java
```diff
@@ -174,8 +174,10 @@ public Integer getPlotNo() {
 		return this.plotNo;
 	}
 
-	public void setPlotNo(final Integer plotNo) {
-		this.plotNo = plotNo;
+	public void setPlotNo(final Object plotNo) {
+		if (plotNo instanceof String) {
```
abatac, line 178: Is it possible to just change the plotNo field's type to just String?

abatac (approved):

PR #1415, opened 2023-01-27
IBP-6116: add audit trail and created_by/updated_by values when saving phenotype

cheryleigh:


src/main/resources/liquibase/crop_changelog/23_1_0.xml
```diff
@@ -64,5 +64,81 @@
...
+	<changeSet author="cheryleigh" id="v23.1.0-6">
+ 		<preConditions onFail="MARK_RAN">
+			<and>
+				<tableExists tableName="phenotype"/>
+				<not>
+					<columnExists tableName="phenotype" columnName="updated_by"/>
+				</not>
+			</and>
+		</preConditions>
+		<comment> Add created_by to phenotype table </comment>
```
abatac, line 92: Fix the comment, shoudl be "Add updated by."

src/main/java/org/generationcp/middleware/service/impl/study/StudyInstanceServiceImpl.java
```diff
@@ -217,6 +221,10 @@ public InstanceObservationData addInstanceObservation(final InstanceObservationD
 			new Phenotype(variableId, instanceObservationData.getValue(), experimentModel);
 		phenotype.setCreatedDate(new Date());
 		phenotype.setUpdatedDate(new Date());
+
```
abatac, line 224: Also consider the phenotypes created from SSA and Dataset Importer in updating the updated/created by fields.

PR #1416, opened 2023-01-31
IBP-6162 enforce germplasm name type uniqueness

mderamos:


src/main/resources/liquibase/crop_changelog/23_1_0.xml
```diff
@@ -64,5 +64,337 @@
...
+
+                DECLARE hasExistingNameType INT DEFAULT FALSE;
+                DECLARE isNameTypeUsedByGermplasm INT DEFAULT TRUE;
+                DECLARE done INT DEFAULT FALSE;
+
+                -- retrieved entries will not be updated
+                DECLARE names_with_duplicates CURSOR FOR
+                    SELECT name.nid as `nameId`, name.gid as `gid`, name.ntype as `nameTypeId`, nameType.ftype as `nameFtype`, nameType.ftable as `nameTable`
+                    FROM names name
+                    INNER JOIN udflds nameType ON nameType.fldno = name.ntype
```
abatac, line 275: Also filter by ftype = 'NAME' to ensure only names are processed:

INNER JOIN udflds nameType ON nameType.fldno = name.ntype AND nameType.ftype = 'NAME'

abatac (changes requested):

src/main/resources/liquibase/crop_changelog/23_1_0.xml
```diff
@@ -64,5 +64,339 @@
...
+
+                DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;
+
+                DECLARE EXIT HANDLER FOR SQLEXCEPTION
+                BEGIN
+                    ROLLBACK;
+                    SIGNAL SQLSTATE '45000' SET MESSAGE_TEXT = 'Error in fixing names with invalid nameTypes.';
+                END;
 
+                OPEN gid_with_invalid_name_types;
```
clarysabel, line 149: @mderamos I am concerned about opening a cursor and iterating case by case, there should be way to simplify it. Not only making it more readable but improving the performance too, like creating a temporary table and identify the different scenarios. 
When writing migrations we should avois as much as possible to iterate when we can run massive update or delete sentences. 
I was doing some analysis on this first SP, I will share with you my ideas, I did not have time to test it and complete it but I want to share it and see if you went into a similar exercise. 
CC @abatac

src/main/resources/liquibase/crop_changelog/23_1_0.xml
```diff
@@ -64,5 +64,199 @@
...
+                        FROM udflds
+                        WHERE name_with_incremented_number.fcode_with_increment = udflds.fcode
+                            AND udflds.ftable = 'NAMES')) toCreate
+                        INNER JOIN udflds ON toCreate.ntype = udflds.fldno;
+
+            -- Update the duplicated names with newly created names
+            UPDATE names n
+                INNER JOIN name_with_incremented_number ON n.nid = name_with_incremented_number.nid
+                LEFT JOIN udflds ON name_with_incremented_number.fcode_with_increment = udflds.fcode AND udflds.ftable = 'NAMES'
+                SET n.ntype = udflds.fldno;
```
abatac, line 244: Add condition here: 

WHERE n.ntype <> udflds.fldno;

To make sure that only names that do not match the updated udflds.fldno will be updated.

src/main/resources/liquibase/crop_changelog/23_1_0.xml
```diff
@@ -64,5 +64,199 @@
...
+
+            -- Cleanup temporary table
+            DROP TABLE name_with_duplicates;
```
abatac, line 247: How's the performance of this liquibase script compared to the cursor approach?

abatac (approved):

PR #1436, opened 2023-03-17
IBP-6188 / IBP-6236 Sample Markers

abatac:


src/main/java/org/generationcp/middleware/domain/genotype/SampleGenotypeSearchRequestDTO.java
```diff
@@ -0,0 +1,181 @@
...
+	/**
+	 * List of columns that will be included and retrieved in the query.
+	 * If no columns are specified, the query will retrieve all columns by default.
+	 */
+	private Set<String> visibleColumns = new HashSet<>();
+
+	/**
+	 * This is used by DataTables to ensure that the Ajax returns from server-side processing requests are drawn in sequence
+	 */
+	private String draw;
```
clarysabel, line 31: is this still required? I remember we stopped using this in some tables so our services are not designed for an specific frontend solution

src/main/java/org/generationcp/middleware/service/api/SampleListService.java
```diff
@@ -66,6 +66,8 @@ SampleList moveSampleList(final Integer sampleListId, final Integer newParentFol
 
 	List<SampleListDTO> getSampleLists(final List<Integer> datasetIds);
 
+	List<SampleListDTO> getSampleListsByStudy(Integer studyId, boolean withGenotypesOnly);
```
clarysabel, line 69: sometimes it gets hard to read boolean params. In this case is not only filtering by study but by another boolean param that tends to make our code dirty. A function searchSampleLists could work better, but we can do it later, no need to change it now

src/main/resources/liquibase/crop_changelog/23_2_0.xml
```diff
@@ -0,0 +1,158 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet id="v23.2.0-1" author="mderamos">
```
clarysabel, line 7: file should refer to sprint 23_3 instead of 23_2, as well as the changesets

PR #1445, opened 2023-04-12
IBP-6283: fix filter of location on crop settings manager

cheryleigh:
- separate location search request DTO for brapi v2


src/main/java/org/generationcp/middleware/dao/location/LocationSearchDAOQuery.java
```diff
@@ -192,11 +192,6 @@ private static void addFilters(final SQLQueryBuilder sqlQueryBuilder, final Loca
       sqlQueryBuilder.setParameter("locationNames", request.getLocationNames());
     }
 
-    if (!CollectionUtils.isEmpty(request.getLocationNames())) {
```
abatac, line 195: Is the locationNames filter not needed?

abatac (approved):

PR #1467, opened 2023-06-26
IBP-5655 Propagate Attribute And Passport Descriptors In Advancing Germplasm

abatac:


lik-83 (approved):

src/test/java/org/generationcp/middleware/api/germplasm/GermplasmAttributeServiceImplIntegrationTest.java
```diff
@@ -30,7 +32,7 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 
-public class GermplasmAttributeServiceImplIntegrationTest  extends IntegrationTestBase {
+public class GermplasmAttributeServiceImplIntegrationTest extends IntegrationTestBase {
```
clarysabel, line 35: @abatac we can create a test to check that the attributes were propagated

PR #1468, opened 2023-06-29
IBP-6357: special handling for BrAPI used in KSU to resolve permission

cheryleigh:


src/main/java/org/generationcp/middleware/dao/PermissionDAO.java
```diff
@@ -113,12 +114,17 @@ public PermissionDAO(final Session session) {
 	}
 
 	public List<PermissionDto> getPermissions(final Integer userId, final String cropName, final Integer programId) {
+		return this.getPermissions(userId, cropName, programId, 0);
+	}
+
+	public List<PermissionDto> getPermissions(final Integer userId, final String cropName, final Integer programId, final int isBrapi) {
```
abatac, line 120: Instead of adding isBrapi flag, you can just create a new method without programId "getPermissions(final Integer userId, final String cropName)". Then if programId is null/not specified, just ignore it in the sql filter.

abatac (changes requested):

abatac (approved):

PR #1478, opened 2023-08-14
IBP-6380-PropagateAttributes

mderamos:


src/main/java/org/generationcp/middleware/api/template/TemplateService.java
```diff
@@ -0,0 +1,20 @@
...
+
+    TemplateDTO saveTemplate(TemplateDTO templateDTO);
+
+    TemplateDTO updateTemplate(TemplateDTO templateDTO);
+
+    void deleteTemplate(Integer templateId);
+
+    List<TemplateDTO> getTemplateDTOsByType(String programUUID, String type);
+
+    Template getTemplateByNameAndProgramUUID(String name, String programUUID);
```
abatac, line 17: Return DTO instead of Hibernate object

src/main/java/org/generationcp/middleware/api/template/TemplateService.java
```diff
@@ -0,0 +1,20 @@
...
+
+    Template getTemplateByIdAndProgramUUID(Integer id, String programUUID);
```
abatac, line 19: Return DTO instead of Hibernate object

src/main/java/org/generationcp/middleware/api/template/TemplateServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+        return this.daoFactory.getTemplateDAO().getTemplateByIdAndProgramUUID(templateId, programUUID);
+    }
+
+    @Override
+    public Template getTemplateByNameAndProgramUUID(final String name, final String programUUID) {
+        return this.daoFactory.getTemplateDAO().getTemplateByNameAndProgramUUID(name, programUUID);
+    }
+
+    @Override
+    public TemplateDTO saveTemplate(TemplateDTO templateDTO) {
```
abatac, line 41: Make sure to run code cleanup to all affected files.

